PR #1, opened 2021-12-31
refactor: make MultiThreaded the default mode

huitseeker:
In multithreaded mode, rocksdb uses its own locking to ensure thread-safe access to the DB:
https://docs.rs/rocksdb/latest/rocksdb/trait.ThreadMode.html

This was the intended use for this API, as evidence by storing the DB reference in `Arc`, but
the DB was opened everywhere in SingleThreaded mode, as by default.

This was a bug that the present PR correcs.


gdanezis (approved):
This is indeed what we intended -- many thanks!

PR #2, opened 2022-01-07
Added multi_get and Iter seek

gdanezis:
This PR augments DBMap with two facilities:
* A `multi_get` allowing multiple keys to be retrieved from the DB at once.
* A `seek` that allows advancing a key-value Iter to a particular key (or the one succeeding it), to support index iteration.

Two tests are added to exercise these functions.

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -307,27 +307,24 @@ where
             .with_big_endian()
             .with_fixint_encoding();
 
-        let mut keys_bytes = Vec::with_capacity(keys.len());
-        for k in keys.iter() {
-            let serialized_key = config.serialize(k)?;
-            // fallible call makes use of iterator hard
-            keys_bytes.push((self.cf(), serialized_key));
-        }
+        let cf = self.cf();
 
-        let results = self
-            .rocksdb
-            .multi_get_cf(keys_bytes.iter().map(|(cf, k)| (cf, k)));
+        let keys_bytes: Result<Vec<_>> = keys
+            .iter()
+            .map(|k| Ok((&cf, config.serialize(k)?)))
+            .collect();
 
-        let mut values_parsed = Vec::with_capacity(keys.len());
-        for value_byte in results {
-            // fallible call makes use of iterator hard
-            values_parsed.push(match value_byte? {
-                Some(data) => Some(bincode::deserialize(&data)?),
-                None => None,
-            });
-        }
+        let results = self.rocksdb.multi_get_cf(keys_bytes?);
 
-        Ok(values_parsed)
+        let values_parsed: Result<Vec<_>> = results
+            .into_iter()
+            .map(|value_byte| match value_byte? {
+                Some(data) => Ok(Some(bincode::deserialize(&data)?)),
+                None => Ok(None),
+            })
+            .collect();
+
+        values_parsed
     }
 }
 
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -300,6 +300,35 @@ where
 
         Values::new(db_iter)
     }
+
+    /// Returns a vector of values corresponding to the keys provided.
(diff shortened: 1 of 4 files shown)


huitseeker (commented):
So, globally I'm OK with the `multi_get`, thanks!

I'm quite wary of the iterator skip here.

I actually had that implemented in the past:
https://github.com/MystenLabs/narwhal/pull/11/commits/b1da3aa855e6813a3391cb32f34708ac6b4eae66

And the notion comes with a few difficulties that I'd like at least to see tested before getting the code in. You'll find hints and examples in tests I wrote in the link above.
- what is the behavior (the iterator's next) if the sought key does not exist?
- what is the range of keys we can further seek to, once we've sought to (e.g.) the median of an Iterator?
- is that the semantics we want, and if so, how do we make this clear to the library's consumer? is `skip_to` better?
- or do we want to emulate a `DoubleEndedOperator`, and if so, is it doable (see comments in link above)?

typed-store/src/rocks/mod.rs
```diff
@@ -300,6 +300,35 @@ where
...
+            .multi_get_cf(keys_bytes.iter().map(|(cf, k)| (cf, k)));
+
+        let mut values_parsed = Vec::with_capacity(keys.len());
+        for value_byte in results {
+            // fallible call makes use of iterator hard
+            values_parsed.push(match value_byte? {
+                Some(data) => Some(bincode::deserialize(&data)?),
+                None => None,
+            });
+        }
```
huitseeker: You're looking for the "collecting into a result" pattern: https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

```suggestion
        results
            .into_iter()
            .map(|res_opt_val| {
                res_opt_val.map_err(|e| e.into()).and_then(|opt_data| {
                    opt_data
                        .map(|data| bincode::deserialize(&data))
                        .transpose()
                        .map_err(|e| e.into())
                })
            })
            .collect()
```

Though I'm not sure the suggestion is more readable than what you are doing, so I'd perhaps just remove the comment, it's actually fine to pattern-match here.

gdanezis: Have a look at how I changed it to use iterators. Its something in the middle, where I use ? within the iterator to make things less magic.

typed-store/src/rocks/tests.rs
```diff
@@ -62,6 +62,88 @@ fn test_get() {
...
+    let result = db.multi_get(&[123, 456, 789]).expect("Failed to multi get");
+
+    assert_eq!(result.len(), 3);
+    assert_eq!(result[0], Some("123".to_string()));
+    assert_eq!(result[1], Some("456".to_string()));
+    assert_eq!(result[2], None);
+}
+
+#[test]
+fn test_seek() {
```
huitseeker: nit 
```suggestion
fn test_skip() {
```

typed-store/src/rocks/tests.rs
```diff
@@ -62,6 +62,88 @@ fn test_get() {
...
+    // The counter intuitive ones
+
+    // Skip to pror of first value
+    let key_vals: Vec<_> = db
+        .iter()
+        .skip_prior_to(&000)
+        .expect("Seek failed")
+        .collect();
+    assert_eq!(key_vals.len(), 0);
+}
```
huitseeker: Nit: I would move the following calls to the next `#[test]`:
```
.skip_prior_to(&999)
.skip_prior_to(&000)
```

I would keep the following, but rename its comment, which does not match the call:
```
.skip_to(&000)
```

huitseeker (approved):
Bit messy, but I'll gladly take it thanks!

typed-store/src/rocks/mod.rs
huitseeker: Very nice job!

gdanezis:
> Bit messy, but I'll gladly take it thanks!

Many thanks!

PR #3, opened 2022-01-19
feat: add a certificate verifier matching an expected pre-shared key

huitseeker:
This PR adds a rustls `(Server|Client)Verifier` accepting specifically correctly self-signed certificates where the signer is an expected signer.

As such, it only uses the certificate chain to make sure the certificate is indeed self-signed, and relies on the caller instantiating the scheme with an expected public key. The set of supported algorithms is limited to two mainstream algorithms, and may or may not be expanded to match those supported by webpki in the future.

The PR also add a KISS generator for such self-signed certificates from a simple `ed25519_dalek` public key, and tests the two in conjunction.

To see an example where those custom verifiers are used, see:
https://github.com/quinn-rs/quinn/blob/main/quinn/examples/insecure_connection.rs

### DIFF
--- a/rccheck/Cargo.toml
+++ b/rccheck/Cargo.toml
@@ -11,9 +11,11 @@ edition = "2021"
 
 [dependencies]
 rustls = { version = "0.20.2", default-features = false, features = ["logging", "dangerous_configuration"] }
+serde = { version = "1.0.133", features = ["derive"]}
 tracing = "0.1.29"
 webpki = { version = "0.22.0", features = ["alloc", "std"] }
 x509-parser = "0.12.0"
 
 [dev-dependencies]
+bincode = "1.3.3"
 rcgen = "0.8.14"
--- a/rccheck/Cargo.toml
+++ b/rccheck/Cargo.toml
@@ -1,6 +1,10 @@
 [package]
 name = "rccheck"
 version = "0.1.0"
+license = "Apache-2.0"
+authors = ["Francois Garillot <francois@mystenlabs.com>"]
+description = "a custom verifier for pre-shared keys"
+repository = "https://github.com/mystenlabs/mysten-infra"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/rccheck/Cargo.toml
+++ b/rccheck/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "rccheck"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rustls = { version = "0.20.2", default-features = false, features = ["logging", "dangerous_configuration"] }
+tracing = "0.1.29"
+webpki = { version = "0.22.0", features = ["alloc", "std"] }
+x509-parser = "0.12.0"
+
+[dev-dependencies]
+rcgen = "0.8.14"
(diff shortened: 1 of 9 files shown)


gdanezis:
Ah ok, this is a new utility right?

rccheck/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "rccheck"
+version = "0.1.0"
+license = "Apache-2.0"
+authors = ["Francois Garillot <francois@mystenlabs.com>"]
+description = "a custom verifier for pre-shared keys"
```
gdanezis: Add a hint this concerns quic/tls? Second, the thing that delayed me understanding this is: as far I understand the pre-shared keys are PUBLIC KEYS, so the client knows the public key of the server, and the server know the public key of the client (initiator). NOT pre-shared secret keys, as the ones used in session resumption in TLS <=1.1. I think this is worth pointing out here.

rccheck/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    fn client_auth_root_subjects(&self) -> Option<rustls::DistinguishedNames> {
+        // We can't guarantee subjects before having seen the cert
+        None
+    }
+
+    fn verify_client_cert(
+        &self,
+        end_entity: &rustls::Certificate,
+        intermediates: &[rustls::Certificate],
```
gdanezis, line 132: Are these certificates on the chain? (ie certs signing the cert?)

rccheck/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            &chain,
+            now,
+        )
+        .map_err(pki_error)
+        .map(|_| ClientCertVerified::assertion())
+    }
+}
+
+impl<'a> ServerCertVerifier for Psk<'a> {
+    fn verify_server_cert(
```
gdanezis, line 169: Here I would add some docs about the high level protocol, and the verification steps withint (with numbers), then inside you can comment with reference to these numbers. Allows the crypto reviewer to first understand if the protocol is correct, and then to understand if the implementation is faithful.

rccheck/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            .verify_is_valid_tls_server_cert(
+                SUPPORTED_SIG_ALGS,
+                &webpki::TlsServerTrustAnchors(&trustroots),
+                &chain,
+                webpki_now,
+            )
+            .map_err(pki_error)
+            .map(|_| cert)?;
+
+        let mut peekable = scts.peekable();
```
gdanezis, line 215: Here we seem to be logging missing info, but not aborting / returning Err. What is the nature of these checks?

gdanezis (approved):
I am far from an expert on the latest TLS tech, we should review by other crypto folks as well.

rccheck/Cargo.toml
huitseeker: Pre-shared keys also correspond to a pattern of identification by public keys, notably in the WPA protocol, but happy to clear any mixup!

rccheck/src/lib.rs
huitseeker: Yes, they are. This is a part of the `rustls` API I do not control

rccheck/src/lib.rs
huitseeker: This is the certificate transparency extension of TLS, which we can't do anything clever about, since we are actually bypassing the certificate chain mechanism (we are only using TLS as a vessel for checking a public key against our knowledge + a PoP).

rccheck/src/lib.rs
huitseeker: Excellent suggestion, thanks!

PR #6, opened 2022-01-21
Adds a TypedStore error type

huitseeker:
TypedStore errors are all coerced into the concrete `anyhow::Error`,  which makes the typed_store crate hard to reuse upstream.

This adds a basic Error type that for now encapsulates errors from either the database or the serializer / deserializer.

### DIFF
--- a/typed-store/src/rocks/errors.rs
+++ b/typed-store/src/rocks/errors.rs
@@ -1,0 +1,19 @@
+// Copyright(C) 2022, Mysten Labs
+// SPDX-License-Identifier: Apache-2.0
+
+use bincode::Error as BincodeError;
+use rocksdb::Error as RocksError;
+use thiserror::Error;
+
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum TypedStoreError {
+    #[error("rocksdb error")]
+    RocksDBError(#[from] RocksError),
+    #[error("(de)serialization error")]
+    SerializationError(#[from] BincodeError),
+    #[error("the column family {0} was not registered with the database")]
+    UnregisteredColumn(String),
+    #[error("a batch operation can't operate across databases")]
+    CrossDBBatch,
+}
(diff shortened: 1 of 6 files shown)


typed-store/src/rocks/errors.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl Display for RocksErrorDef {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        self.message.fmt(formatter)
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Hash, Eq, PartialEq, Debug, Error)]
+pub enum BincodeErrorDef {
```
gdanezis: Here are we tying a little too closely the error reporting with the underlying choice of using bincode? Maybe we could compress some of these cases, unless we believe devs using this lib can take a different action depending on the nature of the serialization/deserialization error?

gdanezis (approved):
This is good to go, and the only comment is whether we want to expose internals of bincode to library users. What if we change our encoding at some point? (Unlikely maybe?).

typed-store/src/rocks/errors.rs
```diff
@@ -0,0 +1,113 @@
...
+use rocksdb::Error as RocksError;
+use serde::{Deserialize, Serialize};
+use std::{fmt, fmt::Display};
+use thiserror::Error;
+
+#[non_exhaustive]
+#[derive(Error, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
+pub enum TypedStoreError {
+    #[error("rocksdb error")]
+    RocksDBError(#[from] RocksErrorDef),
```
oxade: This is pretty cool, thanks

typed-store/src/rocks/errors.rs
huitseeker: Yes, indeed. However, the classical technique to do that (a severly constrained trait object) is hampered by us requiring fancy traits (`Clone, Hash, Eq`) on this type because we put the `Result` directly in `HashMap`s.

in the meantine I've found a worse solution around it, but that's probably something we want to fix on the logn term.

PR #8, opened 2022-01-27
Added clear function in DBMap

oxade:
Mimicking HashMap::clear to clear all items in a map.
We currently have no clean way to clear a map without getting all keys and then batch deleting the keys.
This PR uses the RocksDB recommended solution via `delete_range`

If this solution checks out, will add other small utility functions such is_empty and len() _[although these can currently be inferred from map.iter().count()]_

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -282,6 +282,27 @@ where
         Ok(())
     }
 
+    fn clear(&self) -> Result<(), TypedStoreError> {
+        let config = bincode::DefaultOptions::new()
+            .with_big_endian()
+            .with_fixint_encoding();
+        let num_items = self.iter().count();
+        if num_items == 0 {
+            return Ok(());
+        }
+        let end = config.serialize(&self.iter().last().unwrap().0)?;
+        if num_items > 1 {
+            let start = config.serialize(&self.iter().next().unwrap().0)?;
+            // This deletes all but the last elem
+            let _ = self
+                .rocksdb
+                .delete_range_cf(&self.cf(), start, end.clone())?;
+        }
+        // Finally delete the last elem
+        let _ = self.rocksdb.delete(end)?;
+        Ok(())
+    }
+
     fn iter(&'a self) -> Self::Iterator {
         let mut db_iter = self.rocksdb.raw_iterator_cf(&self.cf());
         db_iter.seek_to_first();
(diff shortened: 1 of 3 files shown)


huitseeker (commented):
The more efficient way to do this is [`drop_cf`](https://docs.rs/rocksdb/0.10.1/rocksdb/struct.DB.html#method.drop_cf).

Look at the column family tests for inspiration:
https://github.com/rust-rocksdb/rust-rocksdb/blob/92b27f5d33c0c713f1d3ec8436b0ae49680e3cca/tests/test_column_family.rs#L94-L106

huitseeker:
Here's the non-atomicity scenario :
- you list the keys with `db.keys()` at the time you prepare your batch,
- somebody inserts new keys K1..KN in the table,
- you execute the batch and delete keys you listed in batch preparation.
- K1..KN are still in the DB.

typed-store/src/rocks/mod.rs
```diff
@@ -282,6 +282,21 @@ where
...
+    }
+
+    fn len(&self) -> Result<usize, Self::Error> {
+        Ok(self.iter().count())
+    }
+
+    fn is_empty(&self) -> Result<bool, Self::Error> {
+        Ok(self.iter().count() == 0)
+    }
+
```
huitseeker: I'd be totally favor of this if we had a way to implement these functions in O(1), instead of O(n). Since we don't these are essentially proxies for the iterator's methods.

Except the iterator's methods:
-  actually are shorter 
- do not imply having to deal with a `Result` which purpose is unclear to me at the moment,
- convey the algorithmic runtime of the function.

huitseeker (commented):
Great job on the column family. 

On the len / `is_emptty`, I don't want to encourage people to call `.len()` and then get surprised by the O(n) runtime. On the other hand `.iter().count()` is right here if they really need it, and conveys the performance tradeoff clearly. Hence I'd be inclined to just remove `len` and `is_empty`.

typed-store/src/rocks/mod.rs
oxade: Hidden O(n) would make me unhappy as a user.
Will remove these and open an issue for designing a constant time soln

huitseeker (approved):
Thanks a bunch!

oxade:
> Thanks a bunch!

Thank you for reviewing and giving feedback!

PR #9, opened 2022-01-28
Iterator extensions

huitseeker:
This adds a convenience implementation of `collectable::TryExtend` for `DBMap` and removes unneeded by-value arguments in batch operations.

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -161,10 +161,10 @@ impl DBBatch {
 impl DBBatch {
     /// Deletes a set of keys given as an iterator
     #[allow(clippy::map_collect_result_unit)] // we don't want a mutable argument
-    pub fn delete_batch<K: Serialize, T: Iterator<Item = K>, V>(
+    pub fn delete_batch<K: Serialize, V>(
         mut self,
         db: &DBMap<K, V>,
-        purged_vals: T,
+        purged_vals: impl IntoIterator<Item = K>,
     ) -> Result<Self, TypedStoreError> {
         if !Arc::ptr_eq(&db.rocksdb, &self.rocksdb) {
             return Err(TypedStoreError::CrossDBBatch);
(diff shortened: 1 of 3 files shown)


typed-store/src/rocks/mod.rs
```diff
@@ -204,15 +204,12 @@ impl DBBatch {
         self.batch.delete_range_cf(&db.cf(), from_buf, to_buf);
         Ok(self)
     }
-}
 
-impl DBBatch {
     /// inserts a range of (key, value) pairs given as an iterator
-    #[allow(clippy::map_collect_result_unit)] // we don't want a mutable argument
-    pub fn insert_batch<K: Serialize, V: Serialize, T: Iterator<Item = (K, V)>>(
+    pub fn insert_batch<J: Borrow<K>, K: Serialize, U: Borrow<V>, V: Serialize>(
```
gdanezis, line 209: If I am now mistaken we can also change the type of pretty much all other functions to only take any type that can be a ref to K instead of K or &K right? Is there a point to that?

gdanezis (approved):
Masterful use of iterators and their strange types, I have to say.

typed-store/src/rocks/mod.rs
huitseeker: > pretty much all other functions

The batch functions (those with an iterator argument) only. I have a PR that demonstrates that in another repo.

> to only take any type that can be a ref to K instead of K or &K right? 

`Borrow` papers over `T` and `&T`, so when you're required to pass an `IntoIterator<Item = (J, U)>` where `J: Borrow<K>, U: Borrow<V>` you can pass an `Iterator<Item = (&K, &V)>`, an `Iterator<Item = (K, &V)>`, an `Iterator<Item = (&K, V)>` or an `Iterator<Item = (K, V)>` at your convenience.

oxade:
Pretty cool

PR #11, opened 2022-01-31
Use bloom filter when checking if key exists

oxade:
RocksDB provides access to it's bloom filter which can be used to check if a key exists in DRAM.
This might save some IO over the current way which uses a potentially wasteful `get` operation

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -241,6 +241,15 @@ where
     type Values = Values<'a, V>;
 
     fn contains_key(&self, key: &K) -> Result<bool, TypedStoreError> {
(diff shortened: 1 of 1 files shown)


huitseeker (changes requested):
1. This does not account for false positives.
2. `KeyMayExist` is an API maintainers added a long time ago, and FB doesn't usually use it in production now. The scenario where it can be used is very limited.

oxade:
> * This does not account for false positives.
> * `KeyMayExist` is an API maintainers added a long time ago, and FB doesn't usually use it in production now. The scenario where it can be used is very limited.

Ah good to know it's not relevant anymore

PR #16, opened 2022-02-02
Add multi-remove and multi-insert convenience fns

### LINKED ISSUE #15: [DBMap] Add atomic multi-insert and multi-remove convenience methods
Multi-insert and multi-remove are current simulated by using batches however convenience functions could wrap these features without exposing the caller to batches.

oxade:
This resolves: https://github.com/MystenLabs/mysten-infra/issues/15

### DIFF
--- a/typed-store/src/traits.rs
+++ b/typed-store/src/traits.rs
@@ -55,7 +55,7 @@ where
     /// Returns a vector of values corresponding to the keys provided.
     fn multi_get(&self, keys: &[K]) -> Result<Vec<Option<V>>, Self::Error>;
 
-    /// Inserts key-value pairs atomically.
+    /// Inserts key-value pairs.
     fn multi_insert<J, U>(
         &self,
         key_val_pairs: impl IntoIterator<Item = (J, U)>,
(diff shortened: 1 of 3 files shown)


typed-store/src/traits.rs
```diff
@@ -54,4 +54,18 @@ where
...
+    ) -> Result<(), Self::Error>
+    where
+        J: Borrow<K>,
+        U: Borrow<V>;
+
+    /// Removes keys atomically.
+    fn multi_remove<J>(&self, keys: impl IntoIterator<Item = J>) -> Result<(), Self::Error>
+    where
+        J: Borrow<K>;
 }
```
huitseeker: So, we both agree that the `rocksDB` way of doing multi-inserts is the most efficient.

But there are many DBs that do not have atomic multi-inserts (e.g. ledgerDB ), so it might b unfair to request one such implementation in the trait. Could we:
- remove the comment on atomicity, or even better mention that atomicity isn't guaranteed, but some implementers may provide it?
- provide a [default implementation](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations) for the multi-insert here, that would serve as a fallback for those DBs without the functionality?

huitseeker (approved):
thanks a bunch! left one extension possibility

typed-store/src/traits.rs
oxade: Right. Will remove the atomicity claim.

PR #17, opened 2022-02-03
fix: fix build of rust-rocksdb with a patch dependency

huitseeker:
rocksdb and librocksdb-sys depend on each other and since a recent upstream commit, rocksdb point at an outdated version of librocksdb-sys, while librocksdb-sys may be bad. This intercepts the versions they query out of crates.io and puts them in agreement: the older version of librocksdb-sys as of a few h ago.

Upstream issue:
https://github.com/rust-rocksdb/rust-rocksdb/issues/602

patch to be removed when it's fixed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,3 +9,7 @@ lto = true
 codegen-units = 1
 lto = true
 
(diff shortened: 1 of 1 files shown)


gdanezis (approved):
Makes sense!

PR #20, opened 2022-02-07
feat: introduce reopen macro to reopen multiple column family

### LINKED ISSUE #13: [typed_store] Add a macro for an iterated variant of reopen
The `typed_store` is a thin typed layer over an (untyped, bit-string-manipulating) DB, aimed at making it look like a `HashMap`.  
The pattern of usage for it is to attach several "tables" for the DB by reopening the DB at various `ColumnFamily` names, leading to the following pattern (which blows up in production):

https://github.com/MystenLabs/mysten-infra/blob/5221b8ff87e18a1bff2ffe930711685d54c04062/typed-store/src/rocks/mod.rs#L61-L66

This is verbose, but cannot be shortened by a function taking in an iterator (the polymorphic nature of reopen would require that iterator to be an [HList](https://okmij.org/ftp/Haskell/HList-ext.pdf). However, this could be a relatively straightforward [declarative macro](https://doc.rust-lang.org/book/ch19-06-macros.html#declarative-macros-with-macro_rules-for-general-metaprogramming).

akichidis:
Resolves #13 

this PR introduces a new macro `reopen`  that acts as a helper to reopen column families and avoid having repeatable code like:

```
let db_cf_1 = DBMap::<u32,u32>::reopen(&rocks, Some("First_CF")).expect("Failed to open storage"); 
let db_cf_2 = DBMap::<u32,u32>::reopen(&rocks, Some("Second_CF")).expect("Failed to open storage"); 
```

I am 100% sure there are other better ways to do it, but that's a way I thought to explore. Honestly, don't know how much more hustle it will save us, but still it's an approach.

The macro basically creates a new function `reopencfs` that handles the column family reopen. The target for me was to make clear the return types of the initialisation function. That was the main challenge for me, to somehow instruct the macro the required column family Key Value types - that's why I ended up on the solution of generating a new function.

Example usage:

```
// call the reopen macro for the required column families 
// and their corresponding Key-Value types
reopen!("cf_1";<String, u32>, "cf_2";<u32, u32>);

// now call the generated function
let (cf_1, cf_2) = reopencfs(&db);
```

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -21,6 +21,37 @@ mod tests;
 type DBRawIteratorMultiThreaded<'a> =
     rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
 
+#[macro_export]
+// It generates a method (reopencfs) to reopen a list of column families.
+// The macro should be called by passing in the column family names and the
+// corresponding DBMap Key-Value types.
+//
+// Example: "cf_1";<String, u32>, "cf_2";<u32, u32>
+macro_rules! reopen {
+    ( $($cf:expr;<$K:ty, $V:ty>),* ) => {
+        // reopencfs calls the methods to open the configured column families.
+        //
+        // # Arguments
+        //
+        // * `db` - a rocks DB struct
+        //
+        // # Examples
+        //
+        // ```
+        // reopen("cf_1";<String, u32>, "cf_2";<u32, u32>);
+        // ....
+        // let (cf_1, cf_2) = reopencfs(&db);
+        // ```
+        fn reopencfs(db: &Arc<rocksdb::DBWithThreadMode<MultiThreaded>>) -> ($(DBMap<$K, $V>),*) {
+            return (
+            $(
+                DBMap::<$K, $V>::reopen(db, Some($cf)).expect(&format!("Cannot open {} CF.", $cf)[..])
+            ),*
+            )
+        }
+    };
+}
+
 /// An interface to a rocksDB database, keyed by a columnfamily
 #[derive(Clone, Debug)]
 pub struct DBMap<K, V> {
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -22,6 +22,7 @@ type DBRawIteratorMultiThreaded<'a> =
     rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
 
 /// a helper macro to reopen multiple column families
+///
 /// # Arguments
 ///
 /// * `db` - a reference to a rocks DB object
@@ -31,18 +32,18 @@ type DBRawIteratorMultiThreaded<'a> =
 ///
 /// # Examples
 ///
-/// We open two different column families and we successfully write to them
+/// We successfully open two different column families.
 /// ```
-/// # #[macro_use] extern crate reopen;
+/// # use typed_store::reopen;
 /// # use typed_store::rocks::*;
 /// # use tempfile::tempdir;
 ///
 /// # fn main() {
-/// # const FIRST_CF: &str = "First_CF";
-/// # const SECOND_CF: &str = "Second_CF";
+/// const FIRST_CF: &str = "First_CF";
+/// const SECOND_CF: &str = "Second_CF";
 ///
 /// /// Create the rocks database reference for the desired column families
-/// let rocks = open_cf(temp_dir().unwrap(), None, &[FIRST_CF, SECOND_CF]).unwrap();
+/// let rocks = open_cf(tempdir().unwrap(), None, &[FIRST_CF, SECOND_CF]).unwrap();
 ///
 /// /// Now simply open all the column families for their expected Key-Value types
 /// let (db_cf_1, db_cf_2) = reopen!(&rocks, FIRST_CF;<i32, String>, SECOND_CF;<i32, String>);
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -21,34 +21,41 @@ mod tests;
 type DBRawIteratorMultiThreaded<'a> =
     rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
 
+/// a helper macro to reopen multiple column families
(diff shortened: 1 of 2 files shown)


typed-store/src/rocks/mod.rs
```diff
@@ -21,6 +21,37 @@ mod tests;
...
+        // ....
+        // let (cf_1, cf_2) = reopencfs(&db);
+        // ```
+        fn reopencfs(db: &Arc<rocksdb::DBWithThreadMode<MultiThreaded>>) -> ($(DBMap<$K, $V>),*) {
+            return (
+            $(
+                DBMap::<$K, $V>::reopen(db, Some($cf)).expect(&format!("Cannot open {} CF.", $cf)[..])
+            ),*
+            )
+        }
```
huitseeker: I like the use of tuples which compels the caller to have a specific structure to its assignments.

typed-store/src/rocks/mod.rs
```diff
@@ -21,6 +21,37 @@ mod tests;
 type DBRawIteratorMultiThreaded<'a> =
     rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
 
+#[macro_export]
+// It generates a method (reopencfs) to reopen a list of column families.
+// The macro should be called by passing in the column family names and the
+// corresponding DBMap Key-Value types.
+//
+// Example: "cf_1";<String, u32>, "cf_2";<u32, u32>
```
huitseeker: It would be great to add a [documentation test](https://doc.rust-lang.org/rustdoc/documentation-tests.html) here.

typed-store/src/rocks/mod.rs
```diff
@@ -21,6 +21,37 @@ mod tests;
...
+        // * `db` - a rocks DB struct
+        //
+        // # Examples
+        //
+        // ```
+        // reopen("cf_1";<String, u32>, "cf_2";<u32, u32>);
+        // ....
+        // let (cf_1, cf_2) = reopencfs(&db);
+        // ```
+        fn reopencfs(db: &Arc<rocksdb::DBWithThreadMode<MultiThreaded>>) -> ($(DBMap<$K, $V>),*) {
```
huitseeker: [C-CASE](https://rust-lang.github.io/api-guidelines/naming.html): you probably want to rename to `reopen_cfs`

typed-store/src/rocks/mod.rs
```diff
@@ -21,6 +21,37 @@ mod tests;
...
     rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
 
+#[macro_export]
+// It generates a method (reopencfs) to reopen a list of column families.
+// The macro should be called by passing in the column family names and the
+// corresponding DBMap Key-Value types.
+//
+// Example: "cf_1";<String, u32>, "cf_2";<u32, u32>
+macro_rules! reopen {
+    ( $($cf:expr;<$K:ty, $V:ty>),* ) => {
```
huitseeker: If you were to do this:
```rust
    ( $db_ref: expr; $($cf:expr;<$K:ty, $V:ty>),* ) => {
```
 would it allow you directly write the returned tuple a couple of lines later?
 ```rust 
$(
    DBMap::<$K, $V>::reopen($db_ref, Some($cf)).expect(&format!("Cannot open {} CF.", $cf)[..])
),
```

Hence the call site would look like:
```rust
let (db_cf_1 db_cf_2) =   reopen!(&rocks; FIRST_CF;<i32, String>, SECOND_CF;<i32, String>);
```

Advantage: I don't have to remember that `reopencfs` exists it's all in 1 place.
Disadvantage: i don't get as fancy a type error (though failure to expand is still a static error).

huitseeker (commented):
Thanks a bunch!

Main thing:  the macro is completely divorced from the reference to the DB, which creates this call pattern in two steps:
```
    let rocks = open_cf(temp_dir(), None, &[FIRST_CF, SECOND_CF]).unwrap();

    reopen!(FIRST_CF;<i32, String>, SECOND_CF;<i32, String>);
```

If you were to take a reference to the DB in the input to the macro, would it allow you to write a macro that executes statements rather than returns a function?

---
Nit: You can manage the status of the linked issue from a GH PR using [these keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

typed-store/src/rocks/mod.rs
akichidis: Absolutely! Wasn't aware that Rust offers "Documentation test" - that is awesome!

typed-store/src/rocks/mod.rs
akichidis: I've followed your recommendation and now there's no need to maintain this method.

typed-store/src/rocks/mod.rs
akichidis: Ah, to be honest that was my original intention but yesterday when I tried it to do it like this, I wasn't able to successfully pass a reference (db) to use together with the repetitive input, that's why I went with the `reopencfs` approach. Tried again today using your recommendation and it works like a charm :-) - I can't even remember what I was doing wrong. Thanks a lot! Now there is no need to call a separate auto-generated method.

typed-store/src/rocks/mod.rs
akichidis: But good point on the C-CASE, I'll keep it in mind for naming conventions.

typed-store/src/rocks/mod.rs
```diff
@@ -21,6 +21,45 @@ mod tests;
...
+///
+/// # fn main() {
+/// const FIRST_CF: &str = "First_CF";
+/// const SECOND_CF: &str = "Second_CF";
+///
+/// /// Create the rocks database reference for the desired column families
+/// let rocks = open_cf(tempdir().unwrap(), None, &[FIRST_CF, SECOND_CF]).unwrap();
+///
+/// /// Now simply open all the column families for their expected Key-Value types
+/// let (db_cf_1, db_cf_2) = reopen!(&rocks, FIRST_CF;<i32, String>, SECOND_CF;<i32, String>);
```
huitseeker: Nit: we could name those `db_map_1` to emphasize the new type, I suppose.

huitseeker (approved):
Tip top! Great work, thanks a bunch!

oxade:
Nice

PR #22, opened 2022-02-08
[refactor] Refactor rccheck in ways that allow a parallel implementation for another scheme

huitseeker:
A crate purporting to do pre-shared PK authentication using TLS certificates needs to do two things:
1. `(kp: KeyPair) -> rustls::Certificate`, (🔑  into 📜  ) 
2. `(cert: rustls::Certfificate, expected_pk: PublicKey) -> rustls::(Client|Server)Verifier` (✅  the 📜  iff if matches the 🔑 )

Before this PR, this crate did part of 2. generically using the "polymorphic" X509 SPKI byte format:
`(cert: rustls::Certfificate, expected_pk: SPKI) -> rustls::(Client|Server)Verifier`
And the rest was implemented specializing the key types to those of ed25519_dalek:
- `(kp: ed25519_dalek::Keypair) -> rustls::Certificate`,
- `(pk: ed25519_dalek::PublicKey) -> SPKI`

This introduces a trait `Certifiable` that would perform this `ed25519_dalek`-specific part generically and folds our `ed25519_dalek` code as just one implementation of it. This allows supporting e.g. ECDSA by implementing that trait (see #23).


rccheck/src/lib.rs
```diff
@@ -1,5 +1,15 @@
 // Copyright(C) 2022, Mysten Labs
 // SPDX-License-Identifier: Apache-2.0
+
+//! THe purpose of this module is to afford the user that already does public key infrastructure on their hosts the ability to perform
```
gdanezis: nit: THe

gdanezis: > already does public key infrastructure

That already knows the correct public keys of other entities? (Doing PKI sounds a little strange)

gdanezis (approved):
Looks good! The docs really helps with the understanding of the purpose and how it works. What about a blog post, and a release with an example on how to use this in TLS? I think that 99% of project have this problem, and looking at the solution, I would be surprised if they get it right?

huitseeker:
@gdanezis An example of how to do this with QUIC is hot on the heels of this. What is this "blog" you speak of?

PR #25, opened 2022-02-23
work in progress

lanvidr:
This provides a library to provide supervision of resilient components that are continuously running.

See [panic issue](https://github.com/MystenLabs/fastnft/issues/382) and [design doc](https://docs.google.com/document/d/1V_HWUdUVblFyNuWC04gnj7wyK-Hi6mmnZwL38VAg5tc/edit?usp=sharing)


lifecycle-component/src/lib.rs
```diff
@@ -0,0 +1,46 @@
+use std::process::Output;
+use tokio::{select, sync::mpsc::{channel, Receiver, Sender}};
+use futures::stream::FuturesUnordered;
+use std::future::Future;
+use std::path::Iter;
+use futures::future::select_all;
+
+// Error is a trait, not a type, so should I use String here instead?
+type IrrecoverableError = dyn std::error::Error;
```
huitseeker: An error type that is easy to convert to is [`anyhow:Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html). The manual `Box<dyn Error>` would work as well, but is more involved.

component/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+use futures::future;
+use tokio::{sync::mpsc::{channel, Sender, Receiver}};
+use tokio::sync::oneshot::{channel as oneshotChannel, Sender as oneshotSender, Receiver as oneshotReceiver};
+use tokio::time::{Duration, timeout};
+
+type IrrecoverableError = std::io::Error;
+type TokioJoinHande = tokio::task::JoinHandle<()>;
+
+pub trait Manageable {
+    fn start(&self, _: Sender<IrrecoverableError>, _: oneshotReceiver<()>) -> future::BoxFuture<'static, ()>;
```
huitseeker: Could [`async-trait`](https://github.com/dtolnay/async-trait) help?

If you look at the [expansion of the macro it generates](https://github.com/dtolnay/async-trait#explanation) it looks a lot like what you have on the right:
https://docs.rs/futures/latest/futures/future/type.BoxFuture.html

-> it should make `Manageable` more .. manageable?

lanvidr: ooh thanks, 👀

component/src/lib.rs
```diff
@@ -33,81 +31,82 @@ pub trait Manageable {
...
+                        let wrapped_handle: tokio::task::JoinHandle<()> =  self.manageable.start(panic_sender, cancel_receiver).await;
+
+                        self.irrecoverable_signal = panic_receiver;
+                        self.cancellation_signal = cancel_sender;
+                        self.join_handle = Some(wrapped_handle);
+
+                        // // restart, continuous passing style
+                        // let next_component = Supervisor::new(self.manageable.clone());
+                        // self = next_component;
+                        // self.spawn().await;
```
lanvidr: This continuous passing does not work because it violates the ownership, but I could be missing something.

component/src/lib.rs
```diff
@@ -33,81 +31,82 @@ pub trait Manageable {
...
         // select statement that listens for the following cases:
         //
-        // panic signal incoming => log, terminate and restart
+        // Irrecoverable signal incoming => log, terminate and restart
         // completion of the task => we're done! return
 
         if let Some(mut handle) = self.join_handle {
             loop {
                 tokio::select! {
-                    Some(message) = self.panic_signal.recv() => {
```
lanvidr: ^ no rules expected this token in macro call

component/src/lib.rs
```diff
@@ -116,46 +115,32 @@ impl<M: Manageable + Send + Clone> Component<M> {
...
 
+
+        // poll (rather than join) the handle for completion for 2s (poll is non-consuming)
+        // and then abort
         if let Some(mut handle) = self.join_handle {
-            // poll (rather than join) the handle for completion for 2s (poll is non-consuming)
-            // and then abort
             loop {
                 tokio::select! {
-                    _ = tokio::time::sleep(Self::GRACE_TIMEOUT) =>{
```
lanvidr: ^ no rules expected this token in macro call

component/src/lib.rs
huitseeker: At this stage, I don't think we need it: the task is de facto restarted when calling `manageable.start` above.

component/src/lib.rs
huitseeker: Yep, that's the handle polling not composing an async expression using a reference. In other words: 
`result = mut handle` should be `result = &mut handle`.

PR #33, opened 2022-03-09
Replace boilerplate serialization

oxade:
@huitseeker 
Can you provide a descriptive comment for the method.
Also we probably need a better name for it.

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -426,8 +426,8 @@ pub fn open_cf<P: AsRef<Path>>(
     Ok(rocksdb)
 }
 
-/// TODO: Good description of why we're doing this
-#[inline(always)]
+/// TODO: Good description of why we're doing this : RocksDB stores keys in BE and has a seek operator on iterators, see https://github.com/facebook/rocksdb/wiki/Iterator#introduction
+#[inline]
 pub fn be_fix_int_ser<S>(t: &S) -> Result<Vec<u8>, TypedStoreError>
 where
     S: ?Sized + serde::Serialize,
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -211,13 +211,10 @@ impl DBBatch {
             return Err(TypedStoreError::CrossDBBatch);
         }
 
-        let config = bincode::DefaultOptions::new()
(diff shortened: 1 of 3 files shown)


asonnino (approved):
LGTM

If I want to use the function `skip_prior_to` of the store with a key of type `(TransactionDigest, SequenceNumber)`, is there a way to tell the store to "sort" those keys by sequence number rather than a weird mix of TransactionDigest and SequenceNumber?

typed-store/src/rocks/mod.rs
```diff
@@ -451,3 +425,16 @@ pub fn open_cf<P: AsRef<Path>>(
     };
     Ok(rocksdb)
 }
+
+/// TODO: Good description of why we're doing this
+#[inline(always)]
```
huitseeker: Please replace with `#[inline]`, the always is not the best idea here, even though it sounds like it should:
http://troubles.md/rust-optimization/#dont-use-inline-always

typed-store/src/rocks/mod.rs
```diff
@@ -451,3 +425,16 @@ pub fn open_cf<P: AsRef<Path>>(
...
+pub fn be_fix_int_ser<S>(t: &S) -> Result<Vec<u8>, TypedStoreError>
```
huitseeker, line 431: How about `serialize_big_endian`, or something more evocative?

typed-store/src/rocks/mod.rs
```diff
@@ -451,3 +425,16 @@ pub fn open_cf<P: AsRef<Path>>(
     };
     Ok(rocksdb)
 }
+
+/// TODO: Good description of why we're doing this
```
huitseeker: The doc on the default Rocks comparator being BE is a bit obscure, and you'd have to look into rust-rocksdb for how this is selected every time, but we can certainly mention in a sentence that "rocksdb stores keys in BE and has a seek operator on iterators": 
https://github.com/facebook/rocksdb/wiki/Iterator#introduction

huitseeker (changes requested):
Thanks, this is much cleaner.
The `#[inline(always)]` will hit compilation times (which are already not super) hard, and the ThinLTO can inline most cases already so I think we should get rid of the always. I left a few other remarks.

huitseeker (approved):

PR #34, opened 2022-03-09
feat: add crate that allows printing the name of an enum variant

huitseeker:
Used for logging cool stuff

### DIFF
--- a/name_variant/examples/test.rs
+++ b/name_variant/examples/test.rs
@@ -1,0 +1,32 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use name_variant::NamedVariant;
+
+#[derive(NamedVariant)]
+enum TestEnum {
+    A,
+    B(),
+    C(i32, i32),
+    D { _name: String, _age: i32 },
+    VariantTest,
+}
+
+fn main() {
+    let x = TestEnum::C(1, 2);
+    assert_eq!(x.variant_name(), "C");
+
+    let x = TestEnum::A;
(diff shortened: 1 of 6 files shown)


name_variant/examples/test.rs
```diff
@@ -0,0 +1,32 @@
...
+    assert_eq!(x.variant_name(), "A");
+
+    let x = TestEnum::B();
+    assert_eq!(x.variant_name(), "B");
+
+    let x = TestEnum::D {
+        _name: "Jane Doe".into(),
+        _age: 30,
+    };
+    assert_eq!(x.variant_name(), "D");
```
velvia: Could we add something that uses x after the assert so we can be sure that variant_name() does not move x and allows x to be used later?

name_variant/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    };
+}
+
+fn match_enum_to_string(name: &Ident, variants: &DataEnum) -> proc_macro2::TokenStream {
+    // the variant dispatch proper
+    let mut match_arms = quote! {};
+    for variant in variants.variants.iter() {
+        let variant_ident = &variant.ident;
+        let fields_in_variant = match &variant.fields {
+            Fields::Unnamed(_) => quote_spanned! {variant.span()=> (..) },
```
velvia: space before `=>` ?

velvia (approved):
Thanks @huitseeker .... this looks awesome!  Much simpler than strum.

name_variant/examples/test.rs
huitseeker: I'm now forcing the receiver, using `(&x).variant_name()` syntax.

PR #36, opened 2022-03-13
Psk improvements

huitseeker:
This fixes `Psk`, from a non-allocating structure (for historical reasons not relevant to our purposes) to a regular, allocating one, which wraps a few bytes. 

As a consequence, the lifetime drops off the description of Psk, it becomes trivial to ferry a Psk across threads and give it to rustls to check client certificates with.


asonnino (approved):
LGTM

PR #37, opened 2022-03-14
[DBMap] Added multi-cf opening options and optimised `contains`

gdanezis:
In the past we only accepted a single options for the full DB, now we also accept spearate options for each CF. And a small optimization for `contains`.

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -275,7 +275,7 @@ where
     fn contains_key(&self, key: &K) -> Result<bool, TypedStoreError> {
         let key_buf = be_fix_int_ser(key)?;
         Ok(self.rocksdb.key_may_exist_cf(&self.cf(), &key_buf)
-            && self.get(key).map(|v| v.is_some())?)
+            && self.rocksdb.get_pinned_cf(&self.cf(), &key_buf)?.is_some())
     }
 
     fn get(&self, key: &K) -> Result<Option<V>, TypedStoreError> {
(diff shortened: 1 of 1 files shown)


typed-store/src/rocks/mod.rs
```diff
@@ -407,7 +420,7 @@ pub fn open_cf<P: AsRef<Path>>(
 
     // Customize CFs
 
-    for cf_key in opt_cfs.iter() {
+    for cf_key in opt_cfs.iter().map(|(name, _)| name) {
```
huitseeker: Nit: 
```suggestion
    for (cf_key, _) in opt_cfs.iter() {
```

typed-store/src/rocks/mod.rs
```diff
@@ -273,7 +273,9 @@ where
     type Values = Values<'a, V>;
 
     fn contains_key(&self, key: &K) -> Result<bool, TypedStoreError> {
-        self.get(key).map(|v| v.is_some())
+        let key_buf = be_fix_int_ser(key)?;
+        Ok(self.rocksdb.key_may_exist_cf(&self.cf(), &key_buf)
```
huitseeker, line 279: You may want to add one line of log saying e.g.
```rust
/// [`rocksdb::DBWithThreadMode::key_may_exist_cf`] can have false positives, but no false negatives. We use it to short-circuit the absent case
```

huitseeker (approved):
Thanks a ton!

stella3d (approved):
@huitseeker already suggested the only change i would, looks good 👍

PR #38, opened 2022-03-15
[feature] add support to remove_all keys atomic operation

akichidis:
While I am working on the issue https://github.com/MystenLabs/narwhal/issues/81 , I found out that the typed-store `Store` doesn't support operations to multi delete (atomically) keys. On this PR I am introducing the following methods to `Store`:
* `write_all`
* `delete_all`
* `read_all`

operating in an atomic way. Left the extension of the `Map` trait with default iterative implementations for follow [up issue](https://github.com/MystenLabs/mysten-infra/issues/40).

### DIFF
--- a/typed-store/src/lib.rs
+++ b/typed-store/src/lib.rs
@@ -8,7 +8,6 @@
     rust_2021_compatibility
 )]
 
-
 use eyre::Result;
 use serde::{de::DeserializeOwned, Serialize};
 use std::{
@@ -66,8 +65,8 @@ where
                         }
                     }
                     StoreCommand::WriteAll(key_values, sender) => {
-                        let response = keyed_db.multi_insert(key_values.iter()
-                            .map(|(k,v)|(k,v)));
+                        let response =
+                            keyed_db.multi_insert(key_values.iter().map(|(k, v)| (k, v)));
 
                         if response.is_ok() {
                             for (key, _) in key_values {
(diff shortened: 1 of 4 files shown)


gdanezis (approved):
Makes sense!

huitseeker (commented):
Here are a few ways this could make the whole crate better, in order of decreasing priority:
- expose `multi_get` and `multi_insert` as well,
- make sure the argument types of `multi_get` are aligned with the correct ones, i.e. the ones in `multi_remove`, & open the associated issue to adapt in downstream repos,
- extend the tests in `store_tests.rs` (or open an issue to do so)
- have default implementations (relying on iteration) for the 3 methods in the trait, making those members optional, or (optionally) marking the trait fields as requiring atomicity.

(feel free to take an arbitrarily small subset of the above)

typed-store/src/lib.rs
```diff
@@ -69,11 +87,28 @@ where
...
+                        // notify the obligations only when the delete was successful
+                        if response.is_ok() {
+                            for key in keys {
+                                if let Some(mut senders) = obligations.remove(&key) {
+                                    while let Some(s) = senders.pop_front() {
+                                        let _ = s.send(Ok(None));
+                                    }
+                                }
+                            }
+                        }
```
akichidis, line 101: I am not sure that there is value doing this (we do similar for the single Delete operation). Based on the implementation of the `notify_read` the obligation:
* either will be notified immediately if the value is available
* or will be notified once it becomes available

in order for an obligator to be notified when a value gets deleted we should have a value already in place, which means that the obligator will have already received a notification via the earlier `write` operation and then get popped up by the queue. 

The other alternative would be to just notify irrespective of whether we do have a successful delete or not - but again don't know what's the value then.

I might be missing a case here, but if we don't see this possible I would lean towards removing this code

huitseeker: The goal of the `notify_read` operation is indeed to implement a post-write callback signal. 

The use case is to make sure you only move on when you have guarantees you'll be able to recover from a process crash through persistence, and in practice we use it as a barrier (i.e. you have X writes to perform and only move on when they've all been done). I've never seen the value of a `notify_read` actually used or inspected, but rather always discarded: the signal is we're *done* with the next write operation on those keys, rather than the details of the written semantics.

Does that make things more clear? I agree that the name `notify_read` and the discipline of actually returning the value when we don't use it are confusing, and that there are probably simplifications to be found there. But I think discharging an obligation on a delete makes sense to me.

huitseeker (approved):
Thank you so much for adding the tests!

typed-store/src/lib.rs
akichidis: Yeah, I guess looking on the code of `notify_read` it seemed to me that the caller will want to get notified _once the data are written by someone (becoming available)_ ....to give an analogy, like a one-off pub/sub with persistence. Ok cool, I'll leave as is then, but it seems that we might need to iterate this in the future , but probably not a priority for now.

PR #39, opened 2022-03-15
Laura/components

lanvidr:
This provides a library that does supervision of resilient components that are continuously running.

See [panic issue](https://github.com/MystenLabs/fastnft/issues/382) and [design doc](https://docs.google.com/document/d/1V_HWUdUVblFyNuWC04gnj7wyK-Hi6mmnZwL38VAg5tc/edit?usp=sharing)


component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+        let (tx_cancellation, tr_cancellation) = oneshotChannel();
+
+        // call the start method
+        let wrapped_handle: JoinHandle = self
+            .manageable
+            .start(tx_irrecoverable, tr_cancellation)
+            .await;
+
+        // reset the supervision handles & channel end points
+        // dropping the old cancellation_signal implicitly sends cancellation by closing the channel
```
lanvidr, line 134: this turns out to be ideal because sending a message on a channel without a receiver that is listening causes an error, so this implicit style frees us of reliance on correct user behavior

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+            manageable: component,
+        }
+    }
+
+    /// Spawn calls the start function of the Manageable component and runs supervision.
+    pub async fn spawn(mut self) -> Result<(), anyhow::Error> {
+        let (tx_irrecoverable, tr_irrecoverable) = channel(10);
+        let (tx_cancellation, tr_cancellation) = oneshotChannel();
+
+        // this assumes the start_fn is populated
```
gdanezis: There is no `start_fn`, maybe say "Manageable::start"?

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+    ) -> tokio::task::JoinHandle<()>; // Note the task is "silent" (returns nothing)
+
+    // The function for cleanup after the task has encountered an irrecoverable error
+    fn handle_irrecoverable(&self, irrecoverable: IrrecoverableError) -> Result<(), anyhow::Error>;
+}
+
+impl<M: Manageable + Send> Supervisor<M> {
+    /// Creates a new supervisor using a Manageable component.
+    pub fn new(component: M) -> Self {
+        let (_, tr_irrecoverable) = channel(10);
```
gdanezis: It seems that on new, the channels assigned to `irrecoverable_signal` and `cancellation_signal` are really dummies, and this is why their respective sending and receiving ends are assigned to `_` (eventually dropped). I think it might be better to make them both options (cancellation_signal already is?) and assign them None. Then once `spawn` is called these may be replaced with Some(...) actual end of channels?

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+        let (tx_cancellation, _) = oneshotChannel();
+        Supervisor {
+            irrecoverable_signal: tr_irrecoverable,
+            cancellation_signal: Some(tx_cancellation),
```
gdanezis: As per my comment above, should this not be assigned None initially?

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Run watches continuously for irrecoverable errors or JoinHandle completion.
+    pub async fn run(mut self) -> Result<(), anyhow::Error> {
+        // select statement that listens for the following cases:
+        //
+        // Irrecoverable signal incoming => log, terminate and restart
+        // completion of the task => already terminated, log and restart
+        //
+        // The handle_irrecoverable is run before the existing task gets
+        // cancelled by restart in the case that an irrecoverable signal
+        // was sent to us. This makes resource cleanup possible.
```
gdanezis, line 102: Probably, here if `self.join_handle.is_none()` we should return an error and not enter the loop, since we are trying to run the task without a proper channel? Or, if this is only intended to be ran by `spawn`, which is guaranteed to initialize the channels, maybe the function `run` should not be `pub` to avoid library user error in calling it before it is ready.

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+               // Poll the JoinHandle<O>
+               _result =  self.join_handle.as_mut().unwrap(), if self.join_handle.is_some() => {
+                    // this could be due to an un-caught panic
+                    // we don't have a user-supplied message to log, so we create a generic one
+                    let message = anyhow!("An unexpected shutdown was observed in a component.");
+                    self.manageable.handle_irrecoverable(message)?;
+                    self.restart().await?;
+                }
+            }
+        }
```
gdanezis, line 119: It seems that both arms of the select, eventually call:
```
self.manageable.handle_irrecoverable(message)?;
self.restart().await?;
```
If the processing in both cases is the same, why not put it in one place after the end of the `select!` statement here. Furthermore this allows the input to `handle_irrecoverable` to be `&mut self` (I think, async borrow checker gods willing), which provides flexibility for the managed task to clean up internal mutable state.

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        // call the start method
+        let wrapped_handle: JoinHandle = self
+            .manageable
+            .start(tx_irrecoverable, tr_cancellation)
+            .await;
+
+        // reset the supervision handles & channel end points
+        // dropping the old cancellation_signal implicitly sends cancellation by closing the channel
+        self.irrecoverable_signal = tr_irrecoverable;
```
gdanezis, line 135: It seems like this `irrecoverable_signal` is used to send at most one message, which then triggers restart, and then the channel is dropped and replaced with a new channel. So maybe here you can have a `oneshotChannel` for `irrecoverable_signal`?

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// 2. A handle_irrecoverable which takes actions on a relaunch due to an irrecoverable error
+/// that happened. It takes the error message that may contain a stack trace and other information
+/// that was sent to the Supervisor via the tx_irrecoverable passed into start.
+#[async_trait]
+pub trait Manageable {
+    // The function that spawns a tokio task
+    async fn start(
+        &self,
+        tx_irrecoverable: Sender<anyhow::Error>,
```
gdanezis, line 52: It seems (from `pub fn new(component: M) -> Self`) that we restart upon the manageable task ending with an error, or receiving and error on `tx_irrecoverable`. My intuition would be that `tx_irrecoverable` is to handle external events? Like for example the another component being restarted, and now needing to restart this component. However we seem to pass in `start` the channel sender `tx_irrecoverable`. And also re-init the channel upon a restart (which means other components cannot rely on it to send external signals).

Is that right? What is the purpose of providing two paths for the managed task to signal failure? Should we instead not provide this to managed task, and keep a long term channel we can provide outsiders to force a restart?

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+        rx_cancellation: oneshotReceiver<()>,
```
gdanezis, line 53: The cancellation channels seems to be assigned and moved around, but never sent or received from? To me this is either a sign that cancellation is a separate concern, and they can be removed. Or that we could do something specific upon cancellation: for example provide a default root future for the tasks started, that listens to cancel and exits if the signal is received?

gdanezis: > dropping the old cancellation_signal implicitly sends cancellation by closing the channel

Ok, now I see this comment, and understand the intention of cancellation. So if I understand correctly, the receiver end will get [TryRecvError](https://docs.rs/tokio/latest/tokio/sync/oneshot/error/enum.TryRecvError.html#) on the channel when this is dropped? This is subtle, so you may want to make an `on_cancelled()` async function in `Manageable` that returns a future that fires when a cancel signal is returned. 

There is a little subtlety here (hence why there is value in the library having that code), in that the rx end of the cancel channel needs to be taken as `&mut`. So a test case for waiting to cancel + other event, then another event happening, then waiting to cancel again, then cancelling, would be welcome.

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+};
+
+pub type IrrecoverableError = anyhow::Error;
+type JoinHandle = tokio::task::JoinHandle<()>;
+
+/// A Supervisor is instantiated to supervise a task that should always be running.
+/// A running supervisor will start a component task, and ensure that it is restarted
+/// if it ever stops.
+pub struct Supervisor<M: Manageable> {
+    irrecoverable_signal: Receiver<IrrecoverableError>,
```
gdanezis, line 22: Some docs on each field (even though they are not public) help us remember what they are for when we maintain.

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use anyhow::anyhow;
+use async_trait::async_trait;
+use tokio::sync::mpsc::{channel, Receiver, Sender};
+use tokio::sync::oneshot::{
+    channel as oneshotChannel, Receiver as oneshotReceiver, Sender as oneshotSender,
+};
+
+pub type IrrecoverableError = anyhow::Error;
+type JoinHandle = tokio::task::JoinHandle<()>;
```
gdanezis, line 12: Should we force the join handle to be of `JoinHandle<Result<(), Error>>` to allow the task to return with an error, and then pass in this error to restart. Of course if it ends without an error we should allow the `Manageable` task to exit.

component/examples/stream_example.rs
```diff
@@ -0,0 +1,144 @@
...
+            let mut buf = [0; 10];
+            let _len = match m_tcp.mock_read(&mut buf) {
+                Ok(_) => {} // process
+                Err(_) => {
+                    let e = anyhow!("missing something required");
+                    tx_irrecoverable
+                        .send(e)
+                        .await
+                        .expect("Could not send irrecoverable signal.");
+                    wait_for_cancellation(rx_cancellation).await;
```
gdanezis, line 84: Why not just return here, and end it all? This relates to my questions below on the purpose of cancel why not just return an Err all the way to the root future of the task, and then get it out at the JoinHandle await?

gdanezis (commented):
Thanks so much for this PR, its very important as we build services around multiple long term service tasks, and I am keen to see it protecting and restarting them. I left lots of comments, some of which can directly be used to improve code, and others can become longer term issues. I think some of them can only an answered when we try to apply this pattern in production, and we can observe what we need. Generalising from no (production) example is very hard.

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+            irrecoverable_signal: tr_irrecoverable,
+            cancellation_signal: Some(tx_cancellation),
+            join_handle: None,
+            manageable: component,
+        }
+    }
+
+    /// Spawn calls the start function of the Manageable component and runs supervision.
+    pub async fn spawn(mut self) -> Result<(), anyhow::Error> {
+        let (tx_irrecoverable, tr_irrecoverable) = channel(10);
```
akichidis: Could that be a `oneshotChannel` as well? If I understand correctly, we want to give some extra "buffer" so the sender so it doesn't block in case we send multiple error messages?

akichidis: If this remains a `channel(10)` , I would advise to use a constant instead of passing directly the value `10`, since this seems to be used on the `restart` method as well.

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+            .start(tx_irrecoverable, tr_cancellation)
+            .await;
+
+        // reset the supervision handles & channel end points
+        // dropping the old cancellation_signal implicitly sends cancellation by closing the channel
+        self.irrecoverable_signal = tr_irrecoverable;
+        self.cancellation_signal = Some(tx_cancellation);
+
+        self.join_handle = Some(wrapped_handle);
+        Ok(())
```
akichidis: Apart from the successful scenario where the `Ok` is returned, what are the other cases that the `restart` method can fail? I couldn't identify any , but again I might be missing something because of my inexperience in Rust. Otherwise I would say that the method can just not return any value and give a signal to the user of it that we don't expect this to fail.

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        self.irrecoverable_signal = tr_irrecoverable;
+        self.cancellation_signal = Some(tx_cancellation);
+        self.join_handle = Some(wrapped_handle);
+
+        self.run().await
+    }
+
+    /// Run watches continuously for irrecoverable errors or JoinHandle completion.
+    pub async fn run(mut self) -> Result<(), anyhow::Error> {
```
akichidis: Does this method need to be public?

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+/// that happened. It takes the error message that may contain a stack trace and other information
+/// that was sent to the Supervisor via the tx_irrecoverable passed into start.
+#[async_trait]
+pub trait Manageable {
+    // The function that spawns a tokio task
+    async fn start(
+        &self,
+        tx_irrecoverable: Sender<anyhow::Error>,
+        rx_cancellation: oneshotReceiver<()>,
+    ) -> tokio::task::JoinHandle<()>; // Note the task is "silent" (returns nothing)
```
asonnino, line 54: Don't we want to know what is wrong here?

component/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+license = "Apache-2.0"
+authors = ["Laura Makdah <laura@mystenlabs.com>"]
+description = "library for resilient component supervision"
+repository = "https://github.com/mystenlabs/mysten-infra"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tokio = { version = "1.15.0", features = ["sync", "macros", "rt", "rt-multi-thread", "signal"] }
```
asonnino: Do we need the "signal" feature?

component/src/lib.rs
akichidis: Since this is an Optional , couldn't you just pass a `None` value instead of instantiating a channel? That will be re-created anyways apart of the `spawn` method.

component/examples/stream_example.rs
```diff
@@ -0,0 +1,144 @@
...
+/// supervisor was received before we return which causes the join handle to complete. If we
+/// were to return immediately, there would be no guarantee the message we send will be received.
+async fn wait_for_cancellation(rx_cancellation: oneshotReceiver<()>) {
+    loop {
+        tokio::select! {
+            _ = rx_cancellation => {
+                println!("terminating component task");
+                break;
+            }
+        }
```
asonnino, line 102: What are we selecting on? It seems there is a single branch

component/src/lib.rs
akichidis: Similarly I think you could make the `irrecoverable_signal` an Optional as well, so you avoid doing any channel initialisations here.

component/src/lib.rs
asonnino: Do you have a quick intuition on why the channel size if set to 10? Is it a channel we expect to be use only few times?
Maybe silly question: Why is this channel used more than once?

component/examples/stream_example.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+impl MockTcpStreamComponent {
+    /// This is a function that should run continuously doing some operation, here we are
+    /// continuously listening on a mocked TCP Stream. This is ultimately the function that we
+    /// are supervising.
+    /// Inside this function we first initialize any state that will be used in this component so
+    /// that the scope is also correctly reset on a restart.
+    ///
+    /// This would be an excellent place to also add a scopeguard with a defer_panic so that if the
```
akichidis, line 64: +1 on the defer panic 👍

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+        // was sent to us. This makes resource cleanup possible.
+        loop {
+            tokio::select! {
+                Some(message) = self.irrecoverable_signal.recv() => {
+                    self.manageable.handle_irrecoverable(message)?;
+                    self.restart().await?;
+                },
+
+               // Poll the JoinHandle<O>
+               _result =  self.join_handle.as_mut().unwrap(), if self.join_handle.is_some() => {
```
asonnino, line 112: Why is this unwrap safe?

asonnino: How about some pattern matching like this:
Ok(result) = self.join_handle.as_mut()

component/examples/stream_example.rs
```diff
@@ -0,0 +1,144 @@
...
+#[tokio::main]
+pub async fn main() -> Result<(), anyhow::Error> {
+    // Create a component
+    let stream_component = MockTcpStreamComponent {};
+
+    // Create a supervisor for the component
+    let supervisor = Supervisor::new(stream_component);
+
+    // Spawn the supervisor to start the component and supervision.
+    let _ = match supervisor.spawn().await {
```
akichidis, line 141: I assume on our applications we'll want to wrap the `supervisor.spawn` on a task itself? Or otherwise on a big tokio select statement where we'll have all our components ?

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+                    // this could be due to an un-caught panic
+                    // we don't have a user-supplied message to log, so we create a generic one
+                    let message = anyhow!("An unexpected shutdown was observed in a component.");
+                    self.manageable.handle_irrecoverable(message)?;
+                    self.restart().await?;
+                }
```
asonnino, line 115: Note that if none of the arms of tokio::select! matches, the macro panics

component/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    async fn restart(&mut self) -> Result<(), anyhow::Error> {
+        // restart
+        let (tx_irrecoverable, tr_irrecoverable) = channel(10);
+        let (tx_cancellation, tr_cancellation) = oneshotChannel();
+
+        // call the start method
+        let wrapped_handle: JoinHandle = self
+            .manageable
+            .start(tx_irrecoverable, tr_cancellation)
```
akichidis, line 130: The fact that we start again the component before the `tx_cancellation` channel drops assumes that the user of the component will not do something as part of the "cancellation" process on their side that can mess up with the component's state and be conflicting with the `start` method (if I am getting this correctly). If that's true, maybe we want to make this explicit (add comments?) to the user so they just purely use the channel to wait and not do any state management.

akichidis (approved):
Please take my comments with a (big) pinch of salt since I am still ramping up in Rust :) 

Great work @laura-makdah! I believe that would be super useful for our products! 

A couple of things:
* you include a great example file of how this could be used (really helps). I haven't seen though any `tests` to confirm the correct behaviour of the component. Could you please include a simple case for it?
* a simple `README` file which can explain with a few words why this is useful and a simple setup of it
* if a component miss-behaves it could eternally restart which could be problematic and even worse go undetected. Maybe this could become configurable with some `max_retries` before it permanently fails - or emit a signal/metric? I see this out of scope for this PR, but something that we can discuss if you consider it useful.

I am approving , but definitely would advise to get input from other more experienced Rust eng than me before merging. Also I believe having at least a simple test case would be important.

component/src/lib.rs
lanvidr: 👍

component/src/lib.rs
lanvidr: it should

component/src/lib.rs
lanvidr: removing `pub`

component/src/lib.rs
lanvidr: Yes, that works! Thanks!

component/src/lib.rs
lanvidr: I thought about this as well, and thought about a case where a user has different sub components, which could each send a message through the channel, and thought it was useful to have a small buffer to accommodate different use cases less restrictively. That may turn out to be never done, as you said it is hard to imagine or predict all the ways this will be used.

component/src/lib.rs
lanvidr: I'm removing the dummy for the cancellation_signal. For the irrecoverable_signal, I see the dummy as a small price to pay for less complexity throughout.

component/src/lib.rs
lanvidr: We also listen for the managed task ending because that handles when a code bug panic occurs, or anything that we didn't write code to handle. And `tx_irrecoverable` is indeed that channel we provide to force a restart. When we restart, the old receiver gets dropped along with the task, and so we initialize a new one for the new task. It is a cleaner / simpler to reset all the state, so that ownership and lifetimes are consistent with a single component.

With your insight above about rearranging things so it is possible to pass mut self to handle_irrecoverable, it becomes a bit less important about which path to restart happens. But the benefit of handling an error via passing a message instead of letting it panic remains in that we can pass specific information about what happened to handle_irrecoverable for logging or fine-tuned handling.

component/src/lib.rs
lanvidr: This is indeed very subtle. The purpose of this channel is to avoid a race condition where we send a message on `tx_irrecoverable` and immediately return. In that situation it is a coin flip whether the first branch or the second branch of the `run` method's select statement will get triggered, since the future could return before the message is received, or vice versa. (I have observed this in my testing as well.) So we use the cancellation as an "ack" that the message we sent was received, and then return out of our function. If we don't do this, often times the message we sent will be lost.

component/src/lib.rs
lanvidr: 👍

component/src/lib.rs
lanvidr: My understanding of our intention for these components is for them to be constantly up and running, and if it were possible to somehow end the component, it would constitute as a Byzantine failure. If that is still true, then sending an error in addition to the one passed in the channel may be a bit redundant.

component/examples/stream_example.rs
lanvidr: This is such a good question that made me stop and question many things. 

The initial ticket was explicit about using a channel to send notice of events, and it seems to be inspired by the CSP model of concurrency that is also used in Golang. 

If we could simply return an error that bubbles back up to the point of entry for these components, then we could add some simple infinite retries in main, so would we need such a library as this one? I am guessing that returning an error back up a call stack is easier said than done, or could lead to the message getting lost along the way?

component/src/lib.rs
lanvidr: Because we don't expect this to return in the ideal world, we don't return anything. If all goes perfectly as intended, which it inevitably won't, this component should run an infinite loop that handles anything that could go wrong. When the unexpected does inevitably occur, we pass the message back via the channel.

component/Cargo.toml
lanvidr: guess not, thanks!

component/examples/stream_example.rs
laura-makdah: right!

component/src/lib.rs
lanvidr: Not silly!
https://github.com/MystenLabs/mysten-infra/pull/39#discussion_r829286749

component/examples/stream_example.rs
lanvidr: this is me being unfamiliar with rust
edit: I don't know how to make this blocking otherwise, do you?

component/src/lib.rs
lanvidr: That sounds like a match statement's behavior. Tokio::select! [allows waiting on multiple async computations and returns when a single computation completes.](https://tokio.rs/tokio/tutorial/select)

component/src/lib.rs
lanvidr: Based on my understanding the `if self.join_handle.is_some()` makes it safe

component/src/lib.rs
lanvidr: nope, changing this!

component/src/lib.rs
lanvidr: 👍

component/src/lib.rs
lanvidr: you're right, removing return type

component/src/lib.rs
lanvidr: Updating the cancellation to None, https://github.com/MystenLabs/mysten-infra/pull/39#discussion_r829288309

component/examples/stream_example.rs
lanvidr: Use of a scope guard is now completely obsolete since we will now pass a &mut self to handle_irrecoverable!

component/examples/stream_example.rs
lanvidr: Once we spawn the task, we don't have to select, join, or do anything with the join handle. The supervisor takes care of that for us, and this function will run until you ctrl + c.

component/src/lib.rs
lanvidr: We start a completely new component that has no shared state whatsoever with the old component.

lanvidr:
@akichidis Thanks for your review! 

> * you include a great example file of how this could be used (really helps). I haven't seen though any `tests` to confirm the correct behaviour of the component. Could you please include a simple case for it?

The example functions as an executable integration test. You can run it with 
```cargo run --example stream_example```
and then ctrl + c to stop the component. 

> * a simple `README` file which can explain with a few words why this is useful and a simple setup of it

Sounds good, I'll add this.

> * if a component miss-behaves it could eternally restart which could be problematic and even worse go undetected. Maybe this could become configurable with some `max_retries` before it permanently fails - or emit a signal/metric? I see this out of scope for this PR, but something that we can discuss if you consider it useful.

This sounds like a great thing to keep in mind as a user of the library. You can wrap the handle_irrecoverable function as a closure and keep track of how many times it was called, you can also put alerts in handle_irrecoverable. I think we would probably want to alert on restarts rather than eventually stop because we don't want to become byzantine.

File: component/src/lib.rs (hunk 23 above)
Comment (akichidis) [reply component/src/lib.rs:130]: Ah you are absolutely right, just saw again that part.

File: component/examples/stream_example.rs (hunk 21 above)
Comment (akichidis) [reply component/examples/stream_example.rs:141]: I might be missing something, but practically to execute the underlying code of the `supervisor.spawn()` we have to call the `await` method which will actually block the current thread execution. So let's assume that I want to create a few supervisor components, to make those execute I have to call the `await` for them ,right? So I imagine something like this:

```
let supervisor_1 = Supervisor::new(stream_component);

supervisor_1.spawn().await;


let supervisor_2 = Supervisor::new(stream_component);

supervisor_2.spawn().await;

```

now, the `supervisor_2.spawn()` won't be able to execute unless the `supervisor_1.spawn().await` has finished. However, the `spawn` method calls the `run()` which is an infinite loop and will not allow us to continue the execution unless it has been exited. Apologies if I am getting this all wrong.

akichidis:
> The example functions as an executable integration test. You can run it with
 cargo run --example stream_example
 and then ctrl + c to stop the component.

Didn't know about `cargo run --example`, that's super useful - will definitely use it! This is not running though as part of the test suite , right? Also as far as I see it doesn't contain asserts, so my concern is that someone won't be able to verify in an automated way that they broke the behaviour in case they refactor the code.

> This sounds like a great thing to keep in mind as a user of the library. You can wrap the handle_irrecoverable function as a closure and keep track of how many times it was called, you can also put alerts in handle_irrecoverable. I think we would probably want to alert on restarts rather than eventually stop because we don't want to become byzantine.

Good point on handling the alerting via the `handle_irrecoverable` 👍  Ok I see the argument about becoming byzantine, I was just thinking that if we stuck on an eternal loop might make us byzantine since practically we could become unresponsive. Meaning that forcing at some point to panic and restart the node could actually help us here to start clean and go back to normal.

component/examples/stream_example.rs
lanvidr: Ah, you would need to call spawn in its own tokio task for each supervisor, we have something like this currently for our authority code. That is a really good question. I missunderstood your original q!

component/examples/stream_example.rs
velvia: It seems the task isn't actually cancelled here?

component/examples/stream_example.rs
velvia: yeah, I was going to say that the easiest API for the user is to pass in an async function that returns Result, and the result is handled automatically.  I was thinking that also I thought "irrecoverable" was about expect's and unwrap's, not returning Errors which are often expected.  It looks like the model here is to explicitly send an error when we want, but I'm just afraid most people will get lazy and not implement that -- thus I think a panic catcher is important.

velvia (commented):
Sorry I didn't take a look before you merged

lanvidr:
@velvia The lazy will be satisfied, off the shelf, it still catches panics and restarts (without needing to send errors).
The cancellation happens implicitly, see prev comment for more details.

PR #41, opened 2022-03-18
Build fixes

huitseeker:
The most important one is we're hitting an instance of https://github.com/rust-rocksdb/rust-rocksdb/issues/609 in CI `--release` builds of downstream packages.


akichidis (approved):
Thanks for the fix! Are we looking to revert once rocksdb is fixed? Or we don't need bzip2 anyways?

huitseeker:
@akichidis I don't think we'll revert: we don't use bzip2, and I don't think we will at our level of required perf.

PR #42, opened 2022-03-20
fix: Clean up the logic of open_cf_descriptors

huitseeker:
The former logic introduced in #37 did not re-open CFs that were not explicitly listed.


asonnino (approved):
LGTM

PR #43, opened 2022-03-20
RFC: introduce a Non-conflicting insert operator

huitseeker:
This :
- demonstrates the use of a merge operator to implement a "weak insert" (insert on empty-or-equal).

This could:
- be used in a try-and-check logic with a get/multi-get to see if a batch of operations have all performed atomically,
- offer a limited substitute to an undefinable CAS operation on the DB.

This could not:
- implement an across-keys all-or-nothing operation, as any merge operator runs independent read-modify-write operations for each key.

TODO:
- [ ] make the operator opt-in rather than by-default,
- [ ] use `pre` crate to make user certify they installed the merge operator they expect at point of use,
- [ ] open the API to let users customize merge operators themselves: 
- ~demo multiplexing of merge operators~

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -8,7 +8,7 @@ mod values;
 use crate::traits::Map;
 use bincode::Options;
 use collectable::TryExtend;
-use rocksdb::{ColumnFamilyDescriptor, DBWithThreadMode, MultiThreaded, WriteBatch};
+use rocksdb::{ColumnFamilyDescriptor, DBWithThreadMode, MergeOperands, MultiThreaded, WriteBatch};
 use serde::{de::DeserializeOwned, Serialize};
 use std::{borrow::Borrow, marker::PhantomData, path::Path, sync::Arc};
 
@@ -132,6 +132,27 @@ impl<K, V> DBMap<K, V> {
         self.rocksdb
             .cf_handle(&self.cf)
             .expect("Map-keying column family should have been checked at DB creation")
+    }
+}
+
+impl<K, V> DBMap<K, V>
+where
+    K: Serialize + DeserializeOwned,
+    V: Serialize + DeserializeOwned,
+{
+    /// Inserts a value at the given key if the key is unset or already equal to the expected value.
+    /// If the key is already set to the value passed as an argument, this is a noop.
+    /// If the key is set to any other value, this does not modify it.
+    ///
+    /// This relies on a precise merge operator being set on the column family: we set it with default options,
+    /// but this may fail if the DBMap's column family is something else.
+    ///
+    pub fn non_conflicting_insert(&self, key: &K, value: &V) -> Result<(), TypedStoreError> {
+        let key_buf = be_fix_int_ser(key)?;
+        let value_buf = bincode::serialize(value)?;
+
+        let _ = self.rocksdb.merge_cf(&self.cf(), &key_buf, &value_buf)?;
+        Ok(())
     }
 }
 
@@ -405,6 +426,29 @@ where
     }
 }
 
+// This is used internally as a merge operator, see:
+// https://github.com/facebook/rocksdb/wiki/Merge-Operator
+pub(crate) fn non_conflicting_insert_merge(
+    _new_key: &[u8],
+    existing_val: Option<&[u8]>,
+    operands: &MergeOperands,
+) -> Option<Vec<u8>> {
+    let nops = operands.len();
+    let mut result: Vec<u8> = Vec::with_capacity(nops);
+    if let Some(v) = existing_val {
+        for e in v {
+            result.push(*e);
+        }
+    } else {
+        for op in operands {
+            for e in op {
+                result.push(*e);
+            }
+        }
+    }
+    Some(result)
+}
+
 /// Opens a database with options, and a number of column families that are created if they do not exist.
 pub fn open_cf<P: AsRef<Path>>(
     path: P,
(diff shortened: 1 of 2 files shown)


typed-store/src/rocks/mod.rs
```diff
@@ -405,15 +447,43 @@ where
...
+pub(crate) fn non_conflicting_insert_merge(
+    _new_key: &[u8],
+    existing_val: Option<&[u8]>,
+    operands: &MergeOperands,
+) -> Option<Vec<u8>> {
+    let nops = operands.len();
+    let mut result: Vec<u8> = Vec::with_capacity(nops);
+    if let Some(v) = existing_val {
+        for e in v {
+            result.push(*e);
```
gdanezis: I would use `Vec::extend` here? 
https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Extend%3CT%3E

typed-store/src/rocks/mod.rs
```diff
@@ -405,15 +447,43 @@ where
...
 
+// This is used internally as a merge operator, see:
+// https://github.com/facebook/rocksdb/wiki/Merge-Operator
+pub(crate) fn non_conflicting_insert_merge(
+    _new_key: &[u8],
+    existing_val: Option<&[u8]>,
+    operands: &MergeOperands,
+) -> Option<Vec<u8>> {
+    let nops = operands.len();
+    let mut result: Vec<u8> = Vec::with_capacity(nops);
```
gdanezis: I think that nops represents the number of operands, not the length of the resulting vec. Maybe here it is worth allocating at least as much as `existing_val` if there is some of it?

typed-store/src/rocks/mod.rs
```diff
@@ -405,15 +447,43 @@ where
...
+    if let Some(v) = existing_val {
+        for e in v {
+            result.push(*e);
+        }
+    } else {
+        for op in operands {
```
gdanezis: I think this currently would write a long string which is the concatenation of all the merges. Is this what is intended here? I would have expected you want to write the very first merge, and then ignore the remaining ones? If so I would change this line to something like
`if let Some(op) = operands.next()`
and then ignore the remaining ones.

You can make a test case that does 2 merges then a read to check if my understanding above is correct.

typed-store/src/rocks/mod.rs
```diff
@@ -405,15 +447,43 @@ where
...
 pub fn open_cf<P: AsRef<Path>>(
     path: P,
     db_options: Option<rocksdb::Options>,
     opt_cfs: &[&str],
 ) -> Result<Arc<rocksdb::DBWithThreadMode<MultiThreaded>>, TypedStoreError> {
     let options = db_options.unwrap_or_default();
-    let column_descriptors: Vec<_> = opt_cfs.iter().map(|name| (*name, &options)).collect();
-    open_cf_opts(path, Some(options.clone()), &column_descriptors[..])
+
+    let mut cf_options = options.clone();
```
gdanezis, line 474: I would allow users of the DB to set this optionally, particularly if there is some magic typing trick to only expose the merge based operations for DBs that have this option on.

gdanezis (commented):
Thanks for looking into merge ops -- this will become crucial down the line. I have a few comments inline, incl one that needs confirmation on the semantics of multiple merges to a non existstant key.

In terms of the motivation stated: how can use this as a substitute for the Read-Modify-Write operations? The issue I see with this is that such RMW sequence involves a read, which would immediately flatten the merge, therefore not getting much benefit from it.

I think the mechanism we could leverage instead is the GetForUpdate in upstream rocksdb:
https://github.com/facebook/rocksdb/wiki/Transactions#guarding-against-read-write-conflicts
which I have not yet seen exposed in the rust wrapper :(

typed-store/src/rocks/mod.rs
huitseeker: Your understanding is correct, but the "long string" behavior only happens in highly concurrent cases, as the tests (passing before the "fix") show.

huitseeker:
> In terms of the motivation stated:

Essentially:
```rust
while db.batch().multi_get(keys_n_vals.map(|(k, v)| k).unwrap() != keys_n_vals.map(|(k, v)| Some(v)).collect::<Vec<_>>() {
    db.batch().non_conflicting_insert_batch(&db, keys_n_vals).write();
}
<do something requiring a weak multi-key lock>
```

Weak here means not pre-empted by another non-conflicting insert op.

It's a poor substitute for an in-memory lock and would likely need to work in conjunction with one of these.

> how can use this as a substitute for the Read-Modify-Write operations? 

I don't understand the question. This *is* a specific RMW op, it does not aim to offer a substitute for the generality of the class of these operations.

> I think the mechanism we could leverage instead is the GetForUpdate in upstream rocksdb:

This is essentially blocked on https://github.com/rust-rocksdb/rust-rocksdb/pull/250

typed-store/src/rocks/mod.rs
huitseeker: Probable. It's in the TODO and will be dealt with before merge.

gdanezis (approved):
Looks good!

typed-store/src/rocks/mod.rs
```diff
@@ -405,15 +447,38 @@ where
...
+pub(crate) fn non_conflicting_insert_merge(
+    _new_key: &[u8],
+    existing_val: Option<&[u8]>,
+    operands: &MergeOperands,
+) -> Option<Vec<u8>> {
+    let nops =
+        operands.iter().map(|v| v.len()).sum::<usize>() + existing_val.map(|_| 1).unwrap_or(0);
+    let mut result: Vec<u8> = Vec::with_capacity(nops);
+    if let Some(v) = existing_val {
+        result.extend(v.iter());
```
velvia, line 459: There's no need to compute nops and allocate a result if there is an existing val, it's wasted.  Also not sure how fast v.iter() is, but it's probably optimized.  However I would do this which should be faster and simpler:

```rust
if let Some(v) = existing_val {
    Some(v.clone())
} else if let Some(op) = operands.into_iter().next() {
    Some(op.clone())
}
```

typed-store/src/rocks/tests.rs
```diff
@@ -99,6 +99,80 @@ fn test_multi_get() {
...
+        .expect("Failed to non-conflicting insert");
+    assert_eq!(
+        Some("123456789".to_string()),
+        db.get(&123456789).expect("Failed to get")
+    );
+
+    // non-conflicting does not work on a set key
+    db.insert(&23456, &"23456".to_string())
+        .expect("Failed to insert");
+    db.non_conflicting_insert(&23456, &"65432".to_string())
```
velvia, line 117: Maybe this is just me but I wish there was a way to return a value indicating whether the write succeeded or not.  To return OK when a write didn't happen seems to be, non ideal from API perspective.

velvia (commented):
Neat.
I can't tell from the RocksDB docs but it seems like you could pass in different merge operator with each C++ API merge, but here in the Rust API you have to set a single merge operator for the entire CF which is too bad.....

typed-store/src/rocks/mod.rs
huitseeker: This is not a total map (`existing_val` and `operands` can both be `None`), and `v`,  `op` are slices, which require a `to_vec()` to allocate besides. But I took some of the feedback into account nonetheless.

PR #45, opened 2022-03-21
New telemetry module: Shared logging/tracing initialization

velvia:
This adds a new module "telemetry" with shared Tokio-tracing subscriber initialization for apps.  Through a config, one can configure for logging/tracing, Tokio console, and JSON output options.  The logging can go to a file optionally instead of stdout.


telemetry/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Getting started is easy.  In your app:
+
+```rust
+  let config = telemetry::TelemetryConfig {
+    service_name: "my_app".into(),
+    ..Default::default()
+  };
+  telemetry::init(config);
+```
```
huitseeker: Could we make this a regularly-compiled snippet? One approach would be a [code example](http://xion.io/post/code/rust-examples.html), which would log to stdout, and show the pretty colors when run.

telemetry/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[dependencies]
+tokio = { version = "1.15.0", features = ["sync", "macros", "rt", "rt-multi-thread"] }
+tracing = { version = "0.1.31", features = ["log"] }
+tracing-appender = "0.2"
+tracing-subscriber = { version = "0.3.9", features = ["time", "registry", "env-filter"] }
+tracing-bunyan-formatter = "0.3"
+tracing-opentelemetry = "0.17"
+opentelemetry = { version = "*", features = ["rt-tokio"] }
+opentelemetry-jaeger = { version = "0.16", features = ["rt-tokio"] }
+console-subscriber = "0.1.3"
```
huitseeker: Unfortunately, all those dependencies will be pulled on each build, which may end up cumbersome. Especially since versioning is a bit brittle.
https://twitter.com/algo_luca/status/1505935455329607687

Could we put a couple of them (open telemetry, jaeger) under a feature flag, and make the dependency optional?

huitseeker (commented):
This is massively useful, thanks!

telemetry/README.md
velvia: I'll add an example for sure.  I'd like to leave an example in the README though just for ease of copy and paste.... but I'm guessing the concern is code rot, in which case we can leave the snippet in the Rustic.

velvia:
@huitseeker example added, and added three different optional features, though jaeger is enabled by default as I think most people will want tracing.

huitseeker (approved):
@velvia Yep, Jaeger default is fine,I just want the ability to unplug stuff I'm not using without code changes, as this is a different crate. Thanks for the PR!

PR #47, opened 2022-03-25
Rename telemetry module to telemetry-subscribers

velvia:
Also add option to set logging level.

The telemetry name was clashing with name of a crate on crates.io, which confuses tools like cargo-deny.


huitseeker (approved):
🙏

PR #48, opened 2022-03-28
[refactor] Use captured identifiers in format strings

huitseeker:
See https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings


akichidis (approved):
Nice, thanks for keeping us up to date!

PR #49, opened 2022-03-29
chore: use a caching install

huitseeker:
https://github.com/baptiste0928/cargo-install caches the binary after compilation, for speed of repeated execution

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -112,9 +112,9 @@ jobs:
           toolchain: nightly
           override: true
       - uses: Swatinem/rust-cache@v1
-      - uses: actions-rs/install@v0.1
+      - name: Install cargo-udeps, and cache the binary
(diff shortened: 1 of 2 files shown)


bmwill (approved):

velvia:
Cool thanks

.github/workflows/rust.yml
```diff
@@ -112,9 +112,9 @@ jobs:
           toolchain: nightly
           override: true
       - uses: Swatinem/rust-cache@v1
-      - uses: actions-rs/install@v0.1
+      - name: Install cargo-udeps, and cache the binary
+        uses: baptiste0928/cargo-install@v1
         with:
           crate: cargo-udeps
-          version: latest
```
bmwill, line 118: Just remembering that we should probably install these binaries in CI using `--locked`

PR #50, opened 2022-03-31
chore: add code coverage to CI

huitseeker:
Tested on my fork: https://app.codecov.io/gh/huitseeker/mysten-infra

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,49 @@
+name: Codecov
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


akichidis (approved):
Woot woot! Nice! 💯 

I couldn't find somewhere any configuration about acceptable coverage thresholds. Do we expect this flow to block our CI if coverage falls under some threshold?

huitseeker:
This does not come with a threshold configuration by default. We can set one through a codecov.yml file: https://docs.codecov.com/docs/common-recipe-list

PR #51, opened 2022-04-04
ci: improve stability and caching

bmwill:
Make a number of improvements to ci similar to what were done here:
https://github.com/MystenLabs/sui/pull/1164


huitseeker (approved):
Thank you so much!

PR #53, opened 2022-04-13
Support for generating chrome trace output

mystenmark:
Can be viewed in the chrome trace viewer via chrome://tracing -> load

Much faster to navigate and explore than Jaeger when you're trying to debug stuff.

image


bmwill (approved):
Looks awesome! The initialization code here is starting to look a bit much due to our happy usage of feature flags. Maybe thats a decision that we'll want to reevaluate at some point in the future but its not relevant to this PR.

oxade:
Thanks @mystenmark!

mystenmark:
@bmwill I think we'll be able to simplify it in the future once we solidify how observability is going to work (e.g. maybe we decide we don't want to use jaeger, or whatever)

PR #60, opened 2022-04-19
anemo: the start of a new p2p networking library

bmwill:
This is the start of the p2p networking library we plan to use in Narwhal and Sui. This is still very much a WIP but I wanted to get this early version out to get some initial feedback on the overall design.

Some areas that are still very much WIP:
- Error handling, right now its non-existent
- Config: There's some hard-coded constants that should be config items
- Builder: building a new network isn't very ergonomic
- Application API: Still need to solidify the message handler interface that applications would need to implement

### DIFF
--- a/anemo/src/common.rs
+++ b/anemo/src/common.rs
@@ -1,0 +1,87 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#[derive(Copy, Clone, Debug, Eq)]
+pub struct PeerId(pub ed25519_dalek::PublicKey);
+
+impl std::hash::Hash for PeerId {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.as_bytes().hash(state);
+    }
+}
+
+impl std::cmp::PartialEq for PeerId {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.as_bytes() == other.0.as_bytes()
+    }
+}
+
+impl std::cmp::PartialOrd for PeerId {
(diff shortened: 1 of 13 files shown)


anemo/src/common.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.as_bytes().hash(state);
+    }
+}
+
+impl std::cmp::PartialEq for PeerId {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.as_bytes() == other.0.as_bytes()
+    }
+}
```
huitseeker, line 17: What's the logic for not deriving these/

anemo/src/crypto.rs
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+static SUPPORTED_SIG_ALGS: &[&webpki::SignatureAlgorithm] = &[&webpki::ED25519];
+
+#[derive(Clone, Debug)]
+pub(crate) struct CertVerifier(pub(crate) String);
```
huitseeker, line 7: You could directly use https://github.com/MystenLabs/mysten-infra/blob/97844c8bb9307d518e879969da527911b5ab9034/rccheck/src/lib.rs#L167-L288 which is just built on a simple generic trait:
https://github.com/MystenLabs/mysten-infra/blob/97844c8bb9307d518e879969da527911b5ab9034/rccheck/src/lib.rs#L45-L58

anemo/src/common.rs
bmwill: The inner type `ed25519_dalek::PublicKey` does not implement these traits so we have to implement them by hand.

PR #65, opened 2022-04-26
[logging] add ability to log to console and files at the same time

longbowlu:
Previously, we can only log to either stdout/stderr or files (apart from other sinks e.g. jaeger). This is not convenient because
1. we want logs in files and pipeline them to a log aggregator aka vector
2. we want to see logs in stdout/stdout because it's much more observable, especially in a container

In this PR, 
1. we add the ability to log to both stdout/stderr and files. Even though logging to file is optional (controlled by `log_file`, ideally we should enable this in all binaries to ramp up observability.
2. in total there are 4 sinks, stdout, stderr, stdout file and stderr, with rules as follow:
```
    // default setting:
    // (after filtering, controlled by env_filter)
    // TRACE -> stdout
    // DEBUG -> stdout
    // INFO  -> stdout, stdout logfile (if enabled)
    // WARN  -> stdout, stdout logfile (if enabled)
    // ERROR -> stderr, stderr logfile (if enabled)
```
3. log files are "{current_dir}/logs/{logfile_prefix}.stdout/stderr.{date}", e.g. `logs/test.stderr.2022-04-26`

Tested locally and things happened in an expected way.

### DIFF
--- a/telemetry_subscribers/src/lib.rs
+++ b/telemetry_subscribers/src/lib.rs
@@ -25,10 +25,12 @@
 use opentelemetry::global;
 #[cfg(feature = "jaeger")]
 use opentelemetry::sdk::propagation::TraceContextPropagator;
+use tracing_subscriber::fmt::writer::BoxMakeWriter;
 
-use tracing::info;
 use tracing::subscriber::set_global_default;
+use tracing::{info, Level};
 use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
+use tracing_subscriber::fmt::writer::MakeWriterExt;
 use tracing_subscriber::{fmt, layer::SubscriberExt, EnvFilter, Registry};
 
 #[cfg(feature = "chrome")]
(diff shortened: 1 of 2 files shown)


huitseeker:
@longbowlu I don't mind this one at all, but it needs a rebase & check.

telemetry_subscribers/src/lib.rs
```diff
@@ -64,14 +68,49 @@ type ChromeGuard = tracing_chrome::FlushGuard;
 #[cfg(not(feature = "chrome"))]
 type ChromeGuard = ();
 
-pub struct TelemetryGuards(WorkerGuard, Option<ChromeGuard>);
+pub struct TelemetryGuards(Vec<WorkerGuard>, Option<ChromeGuard>);
+
+fn get_output(config: &TelemetryConfig) -> (BoxMakeWriter, Vec<WorkerGuard>) {
+    let cur_dir = std::env::current_dir().ok().unwrap().join("logs");
```
velvia, line 74: We should probably let users configure the logging dir via `TelemetryConfig`.

telemetry_subscribers/src/lib.rs
```diff
@@ -64,14 +68,49 @@ type ChromeGuard = tracing_chrome::FlushGuard;
...
 
-fn get_output(config: &TelemetryConfig) -> (NonBlocking, WorkerGuard) {
     if let Some(logfile_prefix) = &config.log_file {
-        let file_appender = tracing_appender::rolling::daily("", logfile_prefix);
-        tracing_appender::non_blocking(file_appender)
+        let (nb_stdout_file, nb_stdout_file_guard) = tracing_appender::non_blocking(
+            tracing_appender::rolling::daily(&cur_dir, format!("{logfile_prefix}.stdout")),
+        );
+        guards.push(nb_stdout_file_guard);
+        let (nb_stderr_file, nb_stderr_file_guard) = tracing_appender::non_blocking(
```
velvia, line 99: Please don't write stderr to a different file.  

It's super important to be able to see the context around errors - to see other INFO, WARN etc around the error.  Having stderr be in a different file makes that really hard to correlate.

Let's just use one file to dump everything in.  People can easily separate out via log level filtering.

telemetry_subscribers/src/lib.rs
```diff
@@ -64,14 +68,49 @@ type ChromeGuard = tracing_chrome::FlushGuard;
...
+    // default setting:
+    // (after filtering, controlled by env_filter)
+    // TRACE -> stdout
+    // DEBUG -> stdout
+    // INFO  -> stdout, stdout logfile (if enabled)
+    // WARN  -> stdout, stdout logfile (if enabled)
+    // ERROR -> stderr, stderr logfile (if enabled)
+
+    let mk_writer = nb_stdout
+        .with_max_level(Level::TRACE)
```
velvia, line 88: 1. Can we move this to later in the path where `config.log_file` is not set?
2. I thought in the description you said that the log levels can be chosen by environment filter, but here we are prescribing which log level... so I don't understand.  This seems to fix the level.  Unless it's a default, but I don't see anything here to interact with env_filter.
3. Trace should not be a default.  it is really really noisy.  The default should be INFO, but if you use env_filter it already defaults to that.

telemetry_subscribers/src/lib.rs
```diff
@@ -64,14 +68,49 @@ type ChromeGuard = tracing_chrome::FlushGuard;
...
+        guards.push(nb_stdout_file_guard);
+        let (nb_stderr_file, nb_stderr_file_guard) = tracing_appender::non_blocking(
+            tracing_appender::rolling::daily(&cur_dir, format!("{logfile_prefix}.stderr")),
+        );
+        guards.push(nb_stderr_file_guard);
+        let mk_writer = mk_writer
+            .and(nb_stderr_file.with_max_level(Level::ERROR))
+            .and(
+                nb_stdout_file
+                    .with_max_level(Level::INFO)
```
velvia, line 107: if anything file logging should default to be more verbose than stdout, because files can easily be bigger and handle more writes, stdout is often used by people to look at what's happening and having too much STDOUT output overwhelms and makes debugging harder.

velvia (changes requested):
1. Single log file output, stdout and stderr should not be separated
2. log levels should not be fixed, should be obtained from env_filter
3. Default should be same log level output for stdout/stderr and log files, maybe a diff setting to choose log file output level?
4. The default should be INFO for stdout/stderr, and maybe more verbose for log files (debug)

telemetry_subscribers/src/lib.rs
mystenmark: agreed on this one

PR #67, opened 2022-05-04
add opening as secondary

oxade:
Allows opening a DB in secondary mode
Code duplication can certainly be reduced
Will revisit

Sample usage https://github.com/MystenLabs/sui/pull/1776

### DIFF
--- a/typed-store/src/rocks/mod.rs
+++ b/typed-store/src/rocks/mod.rs
@@ -456,6 +456,50 @@ pub fn open_cf_opts<P: AsRef<Path>>(
     Ok(rocksdb)
 }
 
+/// Opens a database with options, and a number of column families with individual options that are created if they do not exist.
+pub fn open_cf_opts_secondary<P: AsRef<Path>>(
+    primary_path: P,
+    secondary_path: P,
+    db_options: Option<rocksdb::Options>,
+    opt_cfs: &[(&str, &rocksdb::Options)],
+) -> Result<Arc<rocksdb::DBWithThreadMode<MultiThreaded>>, TypedStoreError> {
+    // Customize database options
+    let mut options = db_options.unwrap_or_default();
+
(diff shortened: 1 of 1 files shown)


huitseeker:
What's the use case for this?

typed-store/src/rocks/mod.rs
```diff
@@ -456,6 +456,50 @@ pub fn open_cf_opts<P: AsRef<Path>>(
...
+    for cf_key in cfs.iter() {
+        if !opt_cfs.contains_key(&cf_key[..]) {
+            opt_cfs.insert(cf_key, &default_rocksdb_options);
+        }
+    }
+
+    let primary_path = primary_path.as_ref().to_path_buf();
+    let mut secondary_path = primary_path.clone();
+    secondary_path.pop();
+    secondary_path.push("/SECONDARY");
```
huitseeker, line 484: I'm OK with the notion of a suffix, but Is there a chance we may clobber another (e.g. existing) secondary path?
You can use this `Into<Option>` technique to add an optional secondary path argument:
https://vidify.org/blog/rust-parameters/#b-with-intooptiont

huitseeker, line 484: No need to add the separator when using `PathBuf::push`: this breaks on non-unixes.
https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push

huitseeker (changes requested):
1. If you want to expose the secondary opening, I think you must expose a variant of [`try_catch_up_with_the_primary`](https://docs.rs/rocksdb/0.15.0/rocksdb/struct.DB.html#method.try_catch_up_with_primary).

2. I'd like to see a test on this:
- open a DB as primary, write one key/value to it,
- open the same DB as secondary with a side path, read the value
- for extra points, have the primary write to the same key twice, and observe the update after `try_catch_up_with_primary`

Requesting changes for at least the separator bug.

oxade:
Not needed any longer

PR #68, opened 2022-05-06
lint: use nexlint to enforce a number of project wide lints

bmwill:


huitseeker (approved):
Tip top!
Note the crate name changes will involve updates on the Sui side.

PR #70, opened 2022-05-09
mysten-network: add common networking utilities

bmwill:
mysten-network: add common networking utilities

    This adds the `mysten-network` crate which can be used to house common
    networking tools across mysten repos. In particular this currently
    includes:
    - A tonic Codec for the bincode format.
    - Utilites and configuration for building tonic servers or clients using
      the `multiaddr` format.

    Supported multiaddr protocol stacks are:
    - dns/tcp/{http,https}
    - ip4/tcp/{http,https}
    - ip6/tcp/{http,https}
    - unix/{http,https}

### DIFF
--- a/crates/mysten-network/src/client.rs
+++ b/crates/mysten-network/src/client.rs
@@ -1,0 +1,167 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    config::Config,
+    multiaddr::{parse_dns, parse_ip4, parse_ip6, parse_unix},
+};
+use anyhow::{anyhow, Context, Result};
+use multiaddr::{Multiaddr, Protocol};
+use std::path::PathBuf;
+use tonic::transport::{Channel, Endpoint, Uri};
+
+pub async fn connect(address: &Multiaddr) -> Result<Channel> {
+    let channel = endpoint_from_multiaddr(address)?.connect().await?;
+    Ok(channel)
+}
+
+pub fn connect_lazy(address: &Multiaddr) -> Result<Channel> {
+    let channel = endpoint_from_multiaddr(address)?.connect_lazy();
+    Ok(channel)
+}
+
+pub(crate) async fn connect_with_config(address: &Multiaddr, config: &Config) -> Result<Channel> {
+    let channel = endpoint_from_multiaddr(address)?
+        .apply_config(config)
+        .connect()
+        .await?;
+    Ok(channel)
+}
+
+pub(crate) fn connect_lazy_with_config(address: &Multiaddr, config: &Config) -> Result<Channel> {
+    let channel = endpoint_from_multiaddr(address)?
+        .apply_config(config)
+        .connect_lazy();
+    Ok(channel)
+}
+
+fn endpoint_from_multiaddr(addr: &Multiaddr) -> Result<MyEndpoint> {
+    let mut iter = addr.iter();
+
+    let channel = match iter.next().ok_or_else(|| anyhow!("address is empty"))? {
+        Protocol::Dns(_) => {
+            let (dns_name, tcp_port, http_or_https) = parse_dns(addr)?;
+            let uri = format!("{http_or_https}://{dns_name}:{tcp_port}");
+            MyEndpoint::try_from_uri(uri)?
+        }
+        Protocol::Ip4(_) => {
+            let (socket_addr, http_or_https) = parse_ip4(addr)?;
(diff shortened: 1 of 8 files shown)


bmwill:
@arun-koshy You may find this of interest as it transparently handles listening on or connecting to Unix Domain Sockets (uds) for doing grpc over IPC.

huitseeker (commented):
Thanks a ton! I think you might have a pkcs8 crate update issue:
https://github.com/MystenLabs/mysten-infra/pull/69

Basically our eddsa25519 version constrains our hashing crates (`Rustcrypto/{traits, hashes}`) versions, which constrain pkcs8. I suspect you're picking a more recent tonic than what we can afford?

See also https://github.com/MystenLabs/mysten-infra/pull/69

bmwill:
@huitseeker it wasn't tonic but an update of ed25519. since we don't have a lockfile checked in the latest semver compatible deps are picked up. I punted on solving the problem by just pinning the ed25519 crate for now.

crates/mysten-network/src/codec.rs
```diff
@@ -0,0 +1,68 @@
...
+            bincode::deserialize_from(buf.reader()).map_err(|e| Status::internal(e.to_string()))?;
+        Ok(Some(item))
+    }
+}
+
+/// A [`Codec`] that implements `application/grpc+bincode` via the serde library.
+#[derive(Debug, Clone)]
+pub struct BincodeCodec<T, U>(PhantomData<(T, U)>);
+
+impl<T, U> Default for BincodeCodec<T, U> {
```
huitseeker, line 45: Does the [trait derivation](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#impl-Default) fall over on the tuple of type arguments?

crates/mysten-network/src/client.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+        unsupported => return Err(anyhow!("unsupported protocol {unsupported}")),
+    };
+
+    Ok(channel)
+}
+
+struct MyEndpoint {
+    endpoint: Endpoint,
+    uds_connector: Option<PathBuf>,
```
huitseeker, line 72: Since this only makes sense in a unix context, would it be better to use [camino's `UTF8PathBuf`](https://github.com/camino-rs/camino)?

crates/mysten-network/src/multiaddr.rs
```diff
@@ -0,0 +1,114 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::{anyhow, Result};
+use multiaddr::{Multiaddr, Protocol};
+use std::{
+    borrow::Cow,
+    net::{IpAddr, SocketAddr},
+};
+
+pub(crate) fn parse_tcp<'a, T: Iterator<Item = Protocol<'a>>>(protocols: &mut T) -> Result<u16> {
```
huitseeker, line 11: Nit: an `impl Iterator<Item = ...>` might lighten the type signature on these `parse_` functions.

huitseeker (commented):
The code looks good but this is sorely lacking in tests. There is at least one test here that you could reuse with only minor modifications:
https://github.com/libp2p/rust-libp2p/blob/22fbce34d5c2c3c7fccaf7c7ce9f80a92406a641/transports/tcp/src/lib.rs#L712-L759

crates/mysten-network/src/codec.rs
bmwill: It falls over due to the generic parameters as the derive puts where clauses for every type which is not what we want.

crates/mysten-network/src/client.rs
bmwill: I could also just leave it as a String as that's the type that the multiaddr library provides

crates/mysten-network/src/codec.rs
huitseeker: Ah, OFC!

crates/mysten-network/src/server.rs
```diff
@@ -0,0 +1,298 @@
...
+    async fn ip6() {
+        let address: Multiaddr = "/ip6/::1/tcp/0/http".parse().unwrap();
+        test_multiaddr(address).await;
+    }
+
+    #[cfg(unix)]
+    #[tokio::test]
+    async fn unix() {
+        // Note that this only works when constructing a multiaddr by hand and not via the
+        // human-readable format
```
huitseeker, line 285: Nice, thanks!

huitseeker (approved):
This is super, thanks a lot!

PR #74, opened 2022-05-13
feat: instrument main RocksDB functions to get tracing output

huitseeker:
This should help visibility into DB errors in dependent crates.


longbowlu (approved):

longbowlu:
merge so we can use to debug

velvia (commented):
Note that this change only dumps out spans.  Default logging won't output spans

PR #79, opened 2022-05-29
tracing: improve default and testing subscribers

bmwill:
Improve the default and testing subscribers to include more information
like the line number as well as outputting span creation and close
events. This also defaults the testing subscriber to Info level instead of Error.

### DIFF
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -26,10 +26,14 @@ use opentelemetry::global;
 #[cfg(feature = "jaeger")]
 use opentelemetry::sdk::propagation::TraceContextPropagator;
 
-use tracing::info;
 use tracing::subscriber::set_global_default;
+use tracing::{info, metadata::LevelFilter};
 use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
-use tracing_subscriber::{fmt, layer::SubscriberExt, EnvFilter, Registry};
+use tracing_subscriber::{
+    fmt::{self, format::FmtSpan},
+    layer::SubscriberExt,
+    EnvFilter, Registry,
+};
 
 #[cfg(feature = "chrome")]
 use tracing_chrome::ChromeLayerBuilder;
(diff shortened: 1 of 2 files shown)


bmwill:
One thing i've wanted is to have filename and line numbers of where logging events occur, unfortunately we end up with it making the logs rather verbose so i'm open to suggestions or if we just shouldn't do filenames

huitseeker (commented):
The best behaviour I've ever seen around this concern is to have a rather "conservative" logging mode applied by default, and a local interface to the executable -using anything from [unix signals](https://github.com/nix-rust/nix/blob/master/src/sys/signal.rs#L682) to an actual full-fleged interactive console-, that allows the operator to turn on a more "liberal" logging mode. 

How far do you think we would be from a primitive verison of that?

bmwill:
> How far do you think we would be from a primitive verison of that?

Not super close, although this is something we can definitely look into building. In the meantime i'll revert the more verbose filename/linenumber for the production logger.

crates/mysten-network/src/server.rs
```diff
@@ -34,12 +38,23 @@ pub struct ServerBuilder {
...
 >;
 
+// type WrapperService =         S: Service<Request<Body>, Response = Response<BoxBody>, Error = Infallible>
+//             + NamedService
+//             + Clone
+//             + Send
+//             + 'static,
+//         S::Future: Send + 'static,
+//         L: Clone,
+
```
huitseeker: Do we want to live this here w/o a comment on what we're trying to achieve?

huitseeker (approved):

crates/mysten-network/src/server.rs
bmwill: Oops that shouldn't be in here. I'll fix before merging

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -184,6 +188,7 @@ pub fn init(config: TelemetryConfig) -> TelemetryGuards {
         // Output to file or to stdout with ANSI colors
         let fmt_layer = fmt::layer()
             .with_ansi(config.log_file.is_none())
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
```
velvia, line 191: We should make this an option to disable, because it's quite verbose and adds lots of information just by itself.

PR #81, opened 2022-06-02
tracing: send logging to stderr instead of stdout by default

bmwill:
In many cases, e.g. cli tools like the wallet, a tool may want to print
out some useful information to stdout. Today we currently log to stdout
which makes it very difficult to distinguish between output we want the
user to actually use and log messages. In order to address this this
patch changes our logging infrastructure to log to stderr by default
instead of stdout.

### DIFF
--- a/crates/telemetry-subscribers/Cargo.toml
+++ b/crates/telemetry-subscribers/Cargo.toml
@@ -8,14 +8,12 @@ repository = "https://github.com/mystenlabs/mysten-infra"
 edition = "2021"
 publish = false
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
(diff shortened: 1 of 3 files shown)


huitseeker (approved):
Code looks good to me, but while I have you here, can I encourage you to add making the `log_file` variable overridable by an environment variable?
See [here](https://mysten-labs.slack.com/archives/C02GD7J9HUM/p1654115825830659) for rationale.

velvia (approved):
Sure.  Though I feel that for CLIs it's better to just log to file by default, as most people will not want to do something extra to redirect stderr.

bmwill:
> Code looks good to me, but while I have you here, can I encourage you to add making the `log_file` variable overridable by an environment variable? See [here](https://mysten-labs.slack.com/archives/C02GD7J9HUM/p1654115825830659) for rationale.

you nerd sniped me...100% of the functionality of tracing can now be controlled via env vars

PR #83, opened 2022-06-22
[mysten_network] add support for metrics reporting

akichidis:
Issue: https://github.com/MystenLabs/narwhal/issues/356

This PR is using several middlewares offered by the `tower-http` crate to allow us monitor endpoint calls for the server. The main one used is the [TraceLayer](https://docs.rs/tower-http/latest/tower_http/trace/struct.TraceLayer.html) It is introducing the `MetricsCallbackProvider` trait which the user can implement and pass when building the server as a way to receive callbacks when the endpoints are called. Special care has been given to also correctly capture the generated [gRPC codes](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc) as the standard http codes don't help in case of errors (as always 200 OK is returned). Also because the endpoint's path is not captured in the `Response` (neither could be extracted from the provided span) that was a limitation when the `on_respond` method was called - and for the purposes of the metrics it is quite important to have the `path` available. For that reason I had to use the combination of the `SetRequestHeaderLayer` and `PropagateHeaderLayer` to set the uri path in the header for each incoming request, and then make sure that this header is copied on the response as well. I consider this a limitation of the `TraceLayer` which could be raised as issue and probably amend so it could make our lives easier.

Why not going with [Tonic interceptor](https://docs.rs/tonic/0.2.0/tonic/struct.Interceptor.html) instead which would probably not even require changes in this library in first place? The interceptor has the limitation that is called only when the request is received and without being able to capture the response (so we couldn't measure latency, response code etc). For that reason it was deemed not appropriate for the purpose.


crates/mysten-network/src/server.rs
```diff
@@ -26,22 +30,42 @@ use tower::{
...
+                SetRequestHeaderLayer<AddPathToHeaderFunction>,
+                Stack<
+                    Either<LoadShedLayer, Identity>,
+                    Stack<Either<GlobalConcurrencyLimitLayer, Identity>, Identity>,
+                >,
+            >,
+        >,
     >,
     Identity,
 >;
```
bmwill: I probably need to get around to making this generic so that we can eliminate this super large type alias

bmwill (approved):

PR #85, opened 2022-06-27
Remove colors from log output when stdout is not tty; remove useless log messages

mystenmark:
Logs in cloudwatch right now are 99% this:

image

This does two things:
* downgrade unhelpful debug-level tracing around DB operations to trace.
* Don't write colored logs to stdout unless on a terminal.

### DIFF
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -17,7 +17,7 @@
 //! - `json` - Bunyan formatter - JSON log output, optional
 //! - `tokio-console` - [Tokio-console](https://github.com/tokio-rs/console) subscriber, optional
 
-use std::env;
+use std::{env, io::stderr};
 use tracing::metadata::LevelFilter;
 use tracing::subscriber::set_global_default;
 use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
@@ -28,6 +28,8 @@ use tracing_subscriber::{
     layer::SubscriberExt,
     EnvFilter, Registry,
 };
+
+use crossterm::tty::IsTty;
 
 /// Configuration for different logging/tracing options
 /// ===
@@ -68,7 +70,7 @@ fn get_output(log_file: Option<String>) -> (NonBlocking, WorkerGuard) {
         let file_appender = tracing_appender::rolling::daily("", logfile_prefix);
         tracing_appender::non_blocking(file_appender)
     } else {
-        tracing_appender::non_blocking(std::io::stderr())
+        tracing_appender::non_blocking(stderr())
     }
 }
 
@@ -218,9 +220,9 @@ impl TelemetryConfig {
                 let json_layer = BunyanFormattingLayer::new(config.service_name, nb_output);
                 Box::new(json_layer)
             } else {
-                // Output to file or to stdout with ANSI colors
+                // Output to file or to stderr with ANSI colors
                 let fmt_layer = fmt::layer()
-                    .with_ansi(config.log_file.is_none())
+                    .with_ansi(config.log_file.is_none() && stderr().is_tty())
                     .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
                     .with_writer(nb_output);
                 Box::new(fmt_layer)
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -197,7 +197,7 @@ impl DBBatch {
     }
 
     /// Consume the batch and write its operations to the database
-    #[instrument(level = "debug", skip_all, err)]
+    #[instrument(level = "trace", skip_all, err)]
     pub fn write(self) -> Result<(), TypedStoreError> {
         self.rocksdb.write(self.batch)?;
         Ok(())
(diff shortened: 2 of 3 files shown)


crates/telemetry-subscribers/src/lib.rs
```diff
@@ -220,7 +222,7 @@ impl TelemetryConfig {
             } else {
                 // Output to file or to stdout with ANSI colors
                 let fmt_layer = fmt::layer()
-                    .with_ansi(config.log_file.is_none())
+                    .with_ansi(config.log_file.is_none() && stdout().is_tty())
```
bmwill: Change this to stderr since we output to stderr by default.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -440,25 +431,38 @@ pub fn open_cf_opts<P: AsRef<Path>>(
...
     opt_cfs: &[(&str, &rocksdb::Options)],
 ) -> Result<Arc<rocksdb::DBWithThreadMode<MultiThreaded>>, TypedStoreError> {
+    println!("BEGIN");
     // Customize database options
     let mut options = db_options.unwrap_or_default();
 
-    let mut opt_cfs: std::collections::HashMap<_, _> = opt_cfs.iter().cloned().collect();
-    let cfs = rocksdb::DBWithThreadMode::<MultiThreaded>::list_cf(&options, &path)
+    let opt_cfs: std::collections::HashMap<_, _> = opt_cfs.iter().cloned().collect();
+    /*
```
bmwill: Did you mean to leave this commented? If so use single line comments. Multi line comments make it difficult to see if code is live or dead when using grep.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -388,7 +380,6 @@ where
     }
 
     /// Convenience method for batch removal
-    #[instrument(level = "debug", skip_all, err)]
```
bmwill, line 391: Instead of removing these we could just lower the level to trace

crates/typed-store/src/rocks/mod.rs
```diff
@@ -471,6 +475,7 @@ pub fn open_cf_opts<P: AsRef<Path>>(
             )?,
         )
     };
+    println!("/whoa");
```
bmwill: Did you mean to leave these added print lines in?

crates/typed-store/src/rocks/mod.rs
mystenmark: wait what the heck this wasn't in the change i pushed....

crates/typed-store/src/rocks/mod.rs
velvia: I agree with @bmwill .... it's useful to instrument key calls, but probably not at debug (but there are lots of things I want to downgrade from debug to trace)

crates/typed-store/src/rocks/mod.rs
```diff
@@ -459,6 +450,8 @@ pub fn open_cf_opts<P: AsRef<Path>>(
     let primary = path.as_ref().to_path_buf();
 
     let rocksdb = {
+        let _span =
+            tracing::debug_span!("rocksdb::DBWithThreadMode::<MultiThreaded>::open_cf_descriptors");
```
velvia: What is this supposed to do?  The span is not used anywhere.  If the intention is just to log instead of trace a span of code, let's use `tracing::trace!` instead.

velvia (commented):
Thanks @mystenmark ... if I understand correctly the intention is to have color output only when the terminal is a real TTY like for dev development.  

Note that `Kubectl logs` can actually display color logs as well, but I think this change might cause all Kubernetes output to be non color, is that right?

Also I think your description was cut off in the PR, you state what the problem is but not the solution... would be good to add a sentence (for search ability) what this PR proposes to do

crates/typed-store/src/rocks/mod.rs
mystenmark: I didn't add or change this code, this was just `cargo fmt`

bmwill: I think this code was added in this PR, I don't see any corresponding '-' lines which you'd expect to see from cargo fmt.

crates/telemetry-subscribers/src/lib.rs
bmwill: @mystenmark did you see this comment?

crates/typed-store/src/rocks/mod.rs
mystenmark: done

crates/telemetry-subscribers/src/lib.rs
mystenmark: yes i just pushed my updates.

crates/typed-store/src/rocks/mod.rs
mystenmark: hmm you're right. my bad again, i started this change in a dirty tree by accident...

mystenmark:
> Note that `Kubectl logs` can actually display color logs as well, but I think this change might cause all Kubernetes output to be non color, is that right?

That is true. If there is a way to have color via kubectl but not in cloudwatch, I'm all ears. Please have a look at the screenshot in the PR description to see what the colors do to cloudwatch.

> Also I think your description was cut off in the PR, you state what the problem is but not the solution... would be good to add a sentence (for search ability) what this PR proposes to do

done

bmwill (approved):

velvia:
> That is true. If there is a way to have color via kubectl but not in cloudwatch, I'm all ears. Please have a look at the screenshot in the PR description to see what the colors do to cloudwatch.

Totally agree we shouldn't have ANSI sequences in CloudWatch.  :)

PR #86, opened 2022-07-03
Convert to Rust 1.62

huitseeker:
+ convert instances of `.then_some()`


crates/mysten-network/src/metrics.rs
```diff
@@ -8,7 +9,8 @@ use tower_http::classify::GrpcFailureClass;
 use tower_http::trace::{OnFailure, OnRequest, OnResponse};
 use tracing::Span;
 
-pub(crate) static GRPC_ENDPOINT_PATH_HEADER: HeaderName = HeaderName::from_static("grpc-path-req");
+pub(crate) static GRPC_ENDPOINT_PATH_HEADER: Lazy<HeaderName> =
+    Lazy::new(|| HeaderName::from_static("grpc-path-req"));
```
bmwill: Whats the reasoning for wrapping this in Lazy when it doesn't actually need to be since its a const function?

huitseeker: `HeaderName::from_static` is not (or no longer) a const function, as you'll be able to notice for yourself thanks to compiler errors when taking https://github.com/MystenLabs/mysten-infra/pull/86/commits/0efc074aeb696aedceaf24de583276d65de9704e and just keeping the version number increase.

bmwill: the latest version of the http crate does have this function as const. the issue is that we aren't checking in lockfiles so your view locally is going to be different from mine, which will be different from CI's.

huitseeker: I added b869c23

bmwill (approved):

PR #89, opened 2022-07-06
Use multi-threaded-cf feature so we can get perf stats

velvia:
This changes the definition of DB to multi-threaded so we can get stats

### DIFF
--- a/crates/typed-store/Cargo.toml
+++ b/crates/typed-store/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 
 [dependencies]
 # deactivation of bzip2 due to https://github.com/rust-rocksdb/rust-rocksdb/issues/609
-rocksdb = { version = "0.18.0", features = ["snappy", "lz4", "zstd", "zlib"], default-features = false }
+rocksdb = { version = "0.18.0", features = ["snappy", "lz4", "zstd", "zlib", "multi-threaded-cf"], default-features = false }
 eyre = "0.6.8"
 serde = { version = "1.0.138", features = ["derive"] }
 bincode = "1.3.3"


huitseeker (commented):
Did you measure the performance impact of the CF's RWLock on the benchmarks we do have (NW bench, Sui micro-bench)?

velvia:
We already use Multithreaded mode, this just changes the definition of a DB type so we can get stats, otherwise our types don't work (without this change, the default type of a DB is `SingleThreaded` which is incompatible with our DB types)

huitseeker:
> We already use Multithreaded mode
Yes. Everywhere in fact:
```
huitseeker@dapple.local➜~/tmp/mysten-infra(main)» rg DBWithThreadMode                                                                                                               [16:02:06]
crates/typed-store/src/rocks/mod.rs
11:use rocksdb::{ColumnFamilyDescriptor, DBWithThreadMode, MultiThreaded, WriteBatch};
23:    rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
69:    pub rocksdb: Arc<rocksdb::DBWithThreadMode<MultiThreaded>>,
113:        db: &Arc<rocksdb::DBWithThreadMode<MultiThreaded>>,
184:    rocksdb: Arc<rocksdb::DBWithThreadMode<MultiThreaded>>,
192:    pub fn new(dbref: &Arc<rocksdb::DBWithThreadMode<MultiThreaded>>) -> Self {
282:        // [`rocksdb::DBWithThreadMode::key_may_exist_cf`] can have false positives,
430:) -> Result<Arc<rocksdb::DBWithThreadMode<MultiThreaded>>, TypedStoreError> {
442:) -> Result<Arc<rocksdb::DBWithThreadMode<MultiThreaded>>, TypedStoreError> {
447:    let cfs = rocksdb::DBWithThreadMode::<MultiThreaded>::list_cf(&options, &path)
465:            rocksdb::DBWithThreadMode::<MultiThreaded>::open_cf_descriptors(
```

> this just changes the definition of a DB type

I'm not sure : [this](https://github.com/rust-rocksdb/rust-rocksdb/blob/64c19df614700d3086bd4caccb9fbfe7a7d3a616/src/db.rs#L297-L298) resets to the DB type we're using so I have zero worries there, but [that](https://github.com/rust-rocksdb/rust-rocksdb/blob/4f695895d0fcd26f09e2934218d9f9351e36e63b/src/column_family.rs#L106-L112) is not covered by what we use.

I expect this new `Arc` to have an impact [here](https://github.com/MystenLabs/mysten-infra/blob/c6dc7a23a40b3517f138d122a76d3bc15f844f67/crates/typed-store/src/rocks/mod.rs#L135-L136) on our side (the deref from `AsColumnFamilyRef` takes the ref on whatever you have for the CF).

velvia:
@huitseeker I don't follow... I don't see `ColumnFamilyRef` used anywhere in the entire mysten-infra repo actually, so I'm not sure where you're expecting an impact.   I see mostly explicit types that specify MultiThreaded everywhere.

huitseeker (approved):
> I don't see ColumnFamilyRef used anywhere in the entire mysten-infra repo actually, 

Actually, never mind, my bad, you were right, we're already [using CFs in the proposed way](https://github.com/MystenLabs/mysten-infra/blob/c6dc7a23a40b3517f138d122a76d3bc15f844f67/crates/typed-store/src/rocks/mod.rs#L134-L138)

This LGTM, sorry for the confusion.

PR #90, opened 2022-07-11
Add opening as secondary

oxade:
Allows opening a DB in secondary mode

### DIFF
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -490,12 +490,9 @@ pub fn open_cf_opts_secondary<P: AsRef<Path>>(
     // Customize database options
     let mut options = db_options.unwrap_or_default();
 
-    // One common issue when running tests on Mac is that the default ulimit is too low,
-    // leading to I/O errors such as "Too many open files". Raising fdlimit to bypass it.
+    fdlimit::raise_fd_limit();
+    // This is a requirement by RocksDB when opening as secondary
     options.set_max_open_files(-1);
-    /* The table cache is locked for updates and this determines the number
-       of shareds, ie 2^10. Increase in case of lock contentions.
-    */
 
     let mut opt_cfs: std::collections::HashMap<_, _> = opt_cfs.iter().cloned().collect();
     let cfs = rocksdb::DBWithThreadMode::<MultiThreaded>::list_cf(&options, &primary_path)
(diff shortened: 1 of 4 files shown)


crates/typed-store/src/rocks/mod.rs
```diff
@@ -474,6 +480,63 @@ pub fn open_cf_opts<P: AsRef<Path>>(
...
+    secondary_path: Option<P>,
+    db_options: Option<rocksdb::Options>,
+    opt_cfs: &[(&str, &rocksdb::Options)],
+) -> Result<Arc<rocksdb::DBWithThreadMode<MultiThreaded>>, TypedStoreError> {
+    // Customize database options
+    let mut options = db_options.unwrap_or_default();
+
+    // One common issue when running tests on Mac is that the default ulimit is too low,
+    // leading to I/O errors such as "Too many open files". Raising fdlimit to bypass it.
+    options.set_max_open_files(-1);
```
huitseeker: We use a one-liner from [`fdlimit`](https://github.com/paritytech/fdlimit) for this in the Sui code base, and could use it here.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -474,6 +480,63 @@ pub fn open_cf_opts<P: AsRef<Path>>(
...
+    }
+
+    let primary_path = primary_path.as_ref().to_path_buf();
+    let secondary_path = secondary_path
+        .map(|q| q.as_ref().to_path_buf())
+        .unwrap_or_else(|| {
+            let mut s = primary_path.clone();
+            s.pop();
+            s.push("SECONDARY");
+            s.as_path().to_path_buf()
```
huitseeker, line 517: You could use the [`std::env!` macro](https://doc.rust-lang.org/std/macro.env.html) to allow overriding this as a one-liner.

huitseeker (commented):
Tip top, I made just a couple suggestions.

crates/typed-store/src/rocks/mod.rs
oxade: That'll cause override at compile time. Why would we want that?

crates/typed-store/src/rocks/mod.rs
oxade: Got it. Will address.

crates/typed-store/src/rocks/mod.rs
huitseeker: 🤦 compile time overrides, not the best idea.
I meant trying to probe the value at runtime. Or a launch argument. Or something other configuration that allows not using the hard coded string.

huitseeker (approved):
Thanks! Nice work, apologies I had missed the `secondary_path` argument.

PR #91, opened 2022-07-12
add iter all function to typed-store

lanvidr:
This is needed for https://github.com/mystenlabs/narwhal/issues/115

### DIFF
--- a/crates/typed-store/src/lib.rs
+++ b/crates/typed-store/src/lib.rs
@@ -38,6 +38,7 @@ pub enum StoreCommand<Key, Value> {
     Read(Key, oneshot::Sender<StoreResult<Option<Value>>>),
     ReadAll(Vec<Key>, oneshot::Sender<StoreResult<Vec<Option<Value>>>>),
     NotifyRead(Key, oneshot::Sender<StoreResult<Option<Value>>>),
+    Iter(oneshot::Sender<HashMap<Key, Value>>),
 }
 
 #[derive(Clone)]
@@ -119,6 +120,10 @@ where
                                 .or_insert_with(VecDeque::new)
                                 .push_back(sender)
                         }
+                    }
+                    StoreCommand::Iter(sender) => {
+                        let response = keyed_db.iter().collect();
+                        let _ = sender.send(response);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


crates/typed-store/src/lib.rs
```diff
@@ -38,6 +38,7 @@ pub enum StoreCommand<Key, Value> {
     Read(Key, oneshot::Sender<StoreResult<Option<Value>>>),
     ReadAll(Vec<Key>, oneshot::Sender<StoreResult<Vec<Option<Value>>>>),
     NotifyRead(Key, oneshot::Sender<StoreResult<Option<Value>>>),
+    Iter(oneshot::Sender<HashMap<Key, Value>>),
```
gdanezis, line 41: The typed store iterator usually returns items in order of keys but a `HashMap` does not. Would it be better to use `BTreeMap` or even a Vec to preserve that order?

gdanezis (approved):
Nice one -- maybe preserve the order?

crates/typed-store/src/lib.rs
lanvidr: For the use case I have, order doesn't matter at all, but this is a good point.

crates/typed-store/src/lib.rs
```diff
@@ -120,6 +121,10 @@ where
                                 .push_back(sender)
                         }
                     }
+                    StoreCommand::Iter(sender) => {
+                        let response = keyed_db.iter().collect();
```
mystenmark, line 125: Wait a second - doesn't this load the whole table into memory?

crates/typed-store/src/lib.rs
```diff
@@ -225,4 +230,14 @@ where
             .await
             .expect("Failed to receive reply to NotifyRead command from store")
     }
+
+    pub async fn iter(&self) -> HashMap<Key, Value> {
```
mystenmark, line 234: shouldn't this return `impl Iterator<Item = (Key, Value)>` - as is this implies loading the whole table into memory.

lanvidr: this was addressed offline

PR #95, opened 2022-07-19
[chore] Updates

huitseeker:
trying to set up a good hash to upgrade NW & Sui to

### DIFF
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -54,7 +54,7 @@ pub struct TelemetryConfig {
     /// If defined, write output to a file starting with this name, ex app.log
     pub log_file: Option<String>,
     /// Log level to set, defaults to info
-    pub log_level: Option<String>,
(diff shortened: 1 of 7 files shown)


bmwill (approved):

kchalkias (approved):

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -54,7 +54,7 @@ pub struct TelemetryConfig {
     /// If defined, write output to a file starting with this name, ex app.log
     pub log_file: Option<String>,
     /// Log level to set, defaults to info
-    pub log_level: Option<String>,
+    pub log_string: Option<String>,
```
velvia, line 57: Well, this is kinda pedantic but I do think `log_level` captures the purpose of this string better than `log_string`.  It is true that it is not a single level, but you can call it `log_level_config`.... it is really specific to levels and not any other log configuration (eg you can't configure logging output directory etc.)

velvia (approved):
Just a minor name comment

PR #96, opened 2022-07-21
tracing: improve the panic handler

bmwill:
Improve our panic handler to do the following:
- Also call the stdlib's panic handler so backtraces aren't swallowed
- flush stdout and stderr
- enable configuring if the panic handler should crash the process


huitseeker (approved):
🤩

Issue #98, opened 2022-07-21
[crypto] Accept an array of public keys in our mTLS implementation


huitseeker:
What we are doing with our `ClientCertVerifier` / `ServerCertVerifier` is accepting one specific key. We can be contacted on the same interface by any of the valid primaries we expect.

We want to extend our "one pre-shared-key" approach only to "an array of pre-shared keys" (the Committee of authorities rather than one specific expected one).

This should:
1. take #97 into account and extend the logic implemented there,
2. take priority over #97 if we have to solve one of these issues rather than both.

PR #100, opened 2022-07-27
[crypto] Create verifier of X.509 `SubjectPublicKeyInfo` against a collection of known PKs.

punwai:
Adds a struct that implements ClientCertVerifier/ServerCertVerifier, which allows the verification of X.509 certificates against a set of pre-shared public keys, as discussed in: #98.

This PR will build up towards implementing mutual TLS (mTLS) [https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/] for connections within Sui and Narwhal.

### DIFF
--- a/crates/rccheck/src/lib.rs
+++ b/crates/rccheck/src/lib.rs
@@ -73,10 +73,10 @@ pub trait Certifiable {
 /// let mut rng = rand::thread_rng();
 /// let keypairs: Vec<_> = (0..10).into_iter().map(|_| ed25519_dalek::Keypair::generate(&mut rng)).collect();
 /// let spkis = keypairs.iter().map(|kp| ed25519_certgen::Ed25519::public_key_to_spki(&kp.public)).collect::<Vec<_>>();
-/// let psket PskSet::from_der(&spkis.iter().map(|spki| &spki[..]).collect::<Vec<_>>()[..]).unwrap();
+/// let pskset = PskSet::from_der(&spkis.iter().map(|spki| &spki[..]).collect::<Vec<_>>()[..]).unwrap();
 /// ```
 ///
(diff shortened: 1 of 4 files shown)


crates/rccheck/src/lib.rs
```diff
@@ -57,6 +62,118 @@ pub trait Certifiable {
...
+        &self,
+        end_entity: &rustls::Certificate,
+        intermediates: &[rustls::Certificate],
+        now: SystemTime,
+    ) -> Result<ClientCertVerified, rustls::Error> {
+        // 1. Check if the key exists in the set of expected keys.
+        let cert = X509Certificate::from_der(&end_entity.0[..])
+            .map_err(|_| rustls::Error::InvalidCertificateEncoding)?;
+        let pk_bytes = cert.1.public_key().raw.to_vec();
+        let spki = Psk::from_der(&pk_bytes).unwrap();
```
kchalkias, line 127: Wondering if it would make sense to have a set of the DER serialized key formats Vs a set of  actual PKs. That would allow for not requiring to extract the key from its DER format saving one deserialization operation.

huitseeker, line 127: Would it be useful to extract this one into a separate function?

huitseeker (approved):
This makes sense, thanks @punwai !

bmwill (approved):
Awesome! Thanks for including a number of tests as well!

crates/rccheck/src/lib.rs
punwai: It does save a lot of deserialization. I wonder why we didn't do this for the original verifier too though - @huitseeker?

punwai: Yes, that would make a lot of sense - adding this in the next PR.

PR #101, opened 2022-07-29
Add a ultility function to convert `MultiAddr` to `SocketAddr`

mwtian:
For https://github.com/MystenLabs/narwhal/issues/575, we need to convert `MultiAddr` to `SocketAddr`, which is accept by the `axum::Server::bind()`. It seems the best place to add this ultility is in `mysten-network`.

### DIFF
--- a/crates/mysten-network/src/multiaddr.rs
+++ b/crates/mysten-network/src/multiaddr.rs
@@ -8,6 +8,21 @@ use std::{
     net::{IpAddr, SocketAddr},
 };
 
+// Converts a /ip{4,6}/-/tcp/- address to SocketAddr.
+pub fn to_socket_addr(addr: &Multiaddr) -> Result<SocketAddr> {
+    let mut iter = addr.iter();
+    let ip = match iter
+        .next()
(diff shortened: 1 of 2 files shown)


crates/mysten-network/src/multiaddr.rs
```diff
@@ -112,3 +127,22 @@ pub(crate) fn parse_unix(address: &Multiaddr) -> Result<(Cow<'_, str>, &'static
...
     Ok((path, http_or_https))
 }
+
+#[cfg(test)]
+mod test {
+    use super::to_socket_addr;
+    use multiaddr::multiaddr;
+
+    #[test]
+    fn test_to_socket_addr() {
```
huitseeker: Would it be worth adding a negative case for e.g. the `Dns` variant of `Protocol`? 

I want to help direct folks to the nearby `client::endpoint_from_multiaddr`, which does a bunch of similar things, and this (name resolution) is where the present function and that one differ.

huitseeker (approved):
LGTM with one question!

crates/mysten-network/src/multiaddr.rs
mwtian: Sounds good. Added a test and a comment for `client::endpoint_from_multiaddr()`

PR #102, opened 2022-07-30
[rccheck] Externally signed x.509 TLS Certificates

punwai:
#### Summary
We need to allow certificates to be signed by other keypairs to allow for the scale-out architecture that we are developing for Narwhal and Sui.

This PR introduces support for externally signed certificates. This is required for the primary/worker split in Narwhal  - we must allow workers to request x.509 certificates from their primaries, which will be used for authentication by other primaries.

### DIFF
--- a/crates/rccheck/src/ed25519_certgen.rs
+++ b/crates/rccheck/src/ed25519_certgen.rs
@@ -12,6 +12,10 @@ use crate::Certifiable;
 #[path = "tests/ed25519_certgen_tests.rs"]
 mod ed25519_certgen_tests;
 
+#[cfg(test)]
+#[path = "tests/ed25519_external_trust_anchor.rs"]
+mod ed25519_external_trust_anchor;
+
 fn dalek_to_keypair_bytes(dalek_kp: ed25519_dalek::Keypair) -> ed25519::KeypairBytes {
     let private = dalek_kp.secret;
     let _public = dalek_kp.public;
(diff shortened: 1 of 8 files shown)


crates/rccheck/src/tests/ed25519_remote_cert.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{ed25519_certgen::Ed25519, test_utils::cert_bytes_to_spki_bytes, *};
+use rcgen::generate_simple_self_signed;
+use rustls::{client::ServerCertVerifier, server::ClientCertVerifier};
+
+#[test]
+fn rc_gen_self_signed_dalek() {
```
punwai: Please check this out and ignore everything else in the PR - is this something that seems viable to do?

crates/rccheck/src/lib.rs
```diff
@@ -473,6 +485,48 @@ fn prepare_for_self_signed<'a>(
...
+) -> Result<CertChainAndRoots<'a>, rustls::Error> {
+    // EE cert must appear first.
+    let cert = webpki::EndEntityCert::try_from(end_entity.0.as_ref()).map_err(pki_error)?;
+
+    let intermediates: Vec<&'a [u8]> = intermediates.iter().map(|cert| cert.0.as_ref()).collect();
+
+    // reinterpret the certificate as a root, materializing the self-signed policy
+    let root_anchor = webpki::TrustAnchor {
+        subject: &[],
+        spki: &root_spki[2..],
```
punwai: Seems... unsafe - but I couldn't find any other interface that allows us to do this without a private key of the spki.

crates/rccheck/src/tests/ed25519_remote_cert.rs
huitseeker: Yes, the script at a high-level looks plausible.

crates/rccheck/src/lib.rs
huitseeker: So, as you noted the fields of this struct are public. This is actually fine (x509 data structures are mostly fitting parsing needs, rather than security functions).

What I don't get is why you need the `[2..]`.

crates/rccheck/src/ed25519_certgen.rs
```diff
@@ -52,6 +56,30 @@ fn gen_certificate(
     Ok(rustls::Certificate(cert_bytes))
 }
 
+/// Signs a CertificateSigningRequest (CSR) and produces a X.509 Certificate.
+fn sign_certificate_signing_request(
+    csr_der: &[u8], // DER-serialized CSR
+    key_pair: (&[u8], &'static SignatureAlgorithm),
```
punwai, line 62: Library requires us pass in the kp...

crates/rccheck/src/ed25519_certgen.rs
```diff
@@ -52,6 +56,30 @@ fn gen_certificate(
...
+) -> Result<rustls::Certificate, anyhow::Error> {
+    let kp = KeyPair::from_der_and_sign_algo(key_pair.0, key_pair.1)?;
+
+    let csr = rcgen::CertificateSigningRequest::from_der(csr_der).map_err(anyhow::Error::new)?;
+
+    let mut cert_params = CertificateParams::new(vec!["localhost".to_string()]);
```
punwai, line 68: Library only allows us to sign CSR with another certificate???

crates/rccheck/src/lib.rs
punwai: I think there is a difference in representation across libraries. One library prefixes with two extra bytes (one byte for the length, another repeats the first byte of the spki)

huitseeker (approved):
Thanks a lot @punwai this looks great

PR #103, opened 2022-07-30
DBMap utils auto derive: reduce boilerplate functions

### LINKED ISSUE #104: [feature] Reduce boilerplate code for DBMap usage
Opening and using DBMaps require lots of boilerplate code. [Example](https://github.com/MystenLabs/sui/blob/main/crates/sui-core/src/authority/authority_store_tables.rs)
Other features like dump also need to be hand written and maintained.

oxade:
_Including everyone whose subsystem touches DBMap, for feedback._

This PR:
* Introduces `#[derive(DBMapUtils)]` which reduces boilerplate code for opening DBMaps and dumping tables.
* Supports DB options as attributes. Example `#[options(optimization = "point_lookup", cache_capacity = 1000000)]`
* Added derive macro and attributes to derive handy utils.
* Makes it easy to expose dumping in various dependent crates with minimal refactoring
* Adds pagination to dump to avoid resource exhaustion

* To ensure debug functions are called only after opening in read only mode, I tried to use preconditions from [pre](https://docs.rs/pre/latest/pre/) but got error: `preconditions are not supported for methods on the stable compiler`

It also basically gets rid of all the functions in [this](https://github.com/MystenLabs/sui/blob/main/crates/sui-core/src/authority/authority_store_tables.rs) file since we derive dump also.

For example this reduces the following [code](https://github.com/MystenLabs/sui/blob/main/crates/sui-storage/src/indexes.rs#L23-L91)

```
pub struct IndexStore {
    /// Index from sui address to transactions initiated by that address.
    transactions_from_addr: DBMap<(SuiAddress, TxSequenceNumber), TransactionDigest>,

    /// Index from sui address to transactions that were sent to that address.
    transactions_to_addr: DBMap<(SuiAddress, TxSequenceNumber), TransactionDigest>,

    /// Index from object id to transactions that used that object id as input.
    transactions_by_input_object_id: DBMap<(ObjectID, TxSequenceNumber), TransactionDigest>,

    /// Index from object id to transactions that modified/created that object id.
    transactions_by_mutated_object_id: DBMap<(ObjectID, TxSequenceNumber), TransactionDigest>,

    /// Index from package id, module and function identifier to transactions that used that moce function call as input.
    transactions_by_move_function:
        DBMap<(ObjectID, String, String, TxSequenceNumber), TransactionDigest>,

    /// This is a map between the transaction digest and its timestamp (UTC timestamp in
    /// **milliseconds** since epoch 1/1/1970). A transaction digest is subjectively time stamped
    /// on a node according to the local machine time, so it varies across nodes.
    /// The timestamping happens when the node sees a txn certificate for the first time.
    timestamps: DBMap<TransactionDigest, u64>,
}

impl IndexStore {
    pub fn open<P: AsRef<Path>>(path: P, db_options: Option<Options>) -> Self {
        let (options, point_lookup) = default_db_options(db_options, Some(1_000_000));

        let db = {
            let path = &path;
            let db_options = Some(options.clone());
            let opt_cfs: &[(&str, &rocksdb::Options)] = &[
                ("transactions_from_addr", &options),
                ("transactions_to_addr", &options),
                ("transactions_by_input_object_id", &options),
                ("transactions_by_mutated_object_id", &options),
                ("transactions_by_move_function", &options),
                ("timestamps", &point_lookup),
            ];
            typed_store::rocks::open_cf_opts(path, db_options, opt_cfs)
        }
        .expect("Cannot open DB.");

        let (
            transactions_from_addr,
            transactions_to_addr,
            transactions_by_input_object_id,
            transactions_by_mutated_object_id,
            transactions_by_move_function,
            timestamps,
        ) = reopen!(
            &db,
            "transactions_from_addr"; <(SuiAddress, TxSequenceNumber), TransactionDigest>,
            "transactions_to_addr"; <(SuiAddress, TxSequenceNumber), TransactionDigest>,
            "transactions_by_input_object_id"; <(ObjectID, TxSequenceNumber), TransactionDigest>,
            "transactions_by_mutated_object_id"; <(ObjectID, TxSequenceNumber), TransactionDigest>,
            "transactions_by_move_function"; <(ObjectID, String, String, TxSequenceNumber), TransactionDigest>,
            "timestamps";<TransactionDigest, u64>
        );

        Self {
            transactions_from_addr,
            transactions_to_addr,
            transactions_by_input_object_id,
            transactions_by_mutated_object_id,
            transactions_by_move_function,
            timestamps,
        }
    }
```
**to** 

```
#[derive(DBMapUtils)]
pub struct IndexStore {
    /// Index from sui address to transactions initiated by that address.
    #[options(cache_capacity = 1000000)]
    transactions_from_addr: DBMap<(SuiAddress, TxSequenceNumber), TransactionDigest>,

    /// Index from sui address to transactions that were sent to that address.
    #[options(cache_capacity = 1000000)]
    transactions_to_addr: DBMap<(SuiAddress, TxSequenceNumber), TransactionDigest>,

    /// Index from object id to transactions that used that object id as input.
    #[options(cache_capacity = 1000000)]
    transactions_by_input_object_id: DBMap<(ObjectID, TxSequenceNumber), TransactionDigest>,

    /// Index from object id to transactions that modified/created that object id.
    #[options(cache_capacity = 1000000)]
    transactions_by_mutated_object_id: DBMap<(ObjectID, TxSequenceNumber), TransactionDigest>,

    /// Index from package id, module and function identifier to transactions that used that moce function call as input.
    #[options(cache_capacity = 1000000)]
    transactions_by_move_function:
        DBMap<(ObjectID, String, String, TxSequenceNumber), TransactionDigest>,

    /// This is a map between the transaction digest and its timestamp (UTC timestamp in
    /// **milliseconds** since epoch 1/1/1970). A transaction digest is subjectively time stamped
    /// on a node according to the local machine time, so it varies across nodes.
    /// The timestamping happens when the node sees a txn certificate for the first time.
    #[options(optimization = "point_lookup", cache_capacity = 1000000)]
    timestamps: DBMap<TransactionDigest, u64>,
}
```

**TODO: more comments and tests**


crates/typed-store-macros-tests/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
```
patrickkuo, line 1: Why put the test in a different crate?

crates/typed-store-macros/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+#[proc_macro_derive(DBMapUtils, attributes(options))]
+pub fn derive_dbmap_utils(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as ItemStruct);
+    let name = &input.ident;
+
+    // Check that the type is DBMap
+    let info = input.fields.iter().map(|f| {
+        // if f.attrs.len() > 1 {
+        //     panic!("Too many attributes. Only `{DB_OPTIONS_ATTR_NAME}` allowed");
+        // }
```
patrickkuo: commented out code

crates/typed-store-macros/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+        use std::path::PathBuf;
+        use rocksdb::Options as RocksDBOptions;
+        use anyhow::anyhow;
+        use std::collections::BTreeMap;
+        use typed_store::rocks;
+        use typed_store::reopen;
+        use typed_store::Map;
+        use rocksdb::MultiThreaded;
+        use std::path::Path as FilePath;
+        use pre::pre;
```
patrickkuo: should avoid generating `use` in proc macro, very likely to cause reimport error when expanded

crates/typed-store-macros/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+
+        const DB_DEFAULT_CF_NAME: &str = "default";
```
patrickkuo: This might cause problem as well if this macro is used more then once in the same file

crates/typed-store-macros/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+                Self {
+                    #(
+                        #field_names,
+                    )*
+                }
+            }
+
+            /// Given a provided `db_options`, add a few default options.
+            /// Returns the default option and the point lookup option.
+            fn adjusted_db_options(
```
patrickkuo: looks like this function don't really need to be in the macro? Can it live in the trait instead?

crates/typed-store-macros/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+                point_lookup.set_memtable_whole_key_filtering(true);
+
+                point_lookup
+            }
+
+            /// List all the tables at this path
+            /// Tables must be opened in read only mode using `open_tables_read_only`
+            /// TODO: use preconditions to ensure call after `open_tables_read_only`
+            // #_precondition_str_tok
+            fn list_tables(path: PathBuf) -> anyhow::Result<Vec<String>> {
```
patrickkuo: ditto

crates/typed-store-macros/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+/// //     bad_field: u32,
+/// // #}
+/// ```
+#[proc_macro_derive(DBMapUtils, attributes(options))]
+pub fn derive_dbmap_utils(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as ItemStruct);
+    let name = &input.ident;
+
+    // Check that the type is DBMap
+    let info = input.fields.iter().map(|f| {
```
patrickkuo: will be more readable if we extract this to `fn extract_info` ?

patrickkuo (changes requested):

crates/typed-store-macros-tests/Cargo.toml
oxade: I tried to put them in same crate but kept getting this error
`error: can't use a procedural macro from the same crate that defines it`
The solution from my research was to separate it out

patrickkuo (approved):
LGTM

PR #109, opened 2022-08-04
Set default total memtable size to 512MB per Rocksdb

mwtian:
The goal is to limit total memtable bytes usage per Rocksdb. Assuming most of memtable memory usages is in 1 or a few dbs, this should reduce per-process memory usage as well. 

I have not verified that this change reduces the maximum total size of memtables in `stress` benchmark though. In fact I'm not sure how to build Sui with this change without merging it.

Related to https://github.com/MystenLabs/narwhal/issues/671

### DIFF
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -316,7 +316,7 @@ where
     fn clear(&self) -> Result<(), TypedStoreError> {
         let _ = self.rocksdb.drop_cf(&self.cf);
         self.rocksdb
-            .create_cf(self.cf.clone(), &rocksdb::Options::default())?;
+            .create_cf(self.cf.clone(), &default_rocksdb_options())?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


crates/typed-store/src/rocks/mod.rs
```diff
@@ -427,14 +427,22 @@ where
     }
 }
 
+/// Creates a default Rocksdb option.
+pub fn default_rocksdb_options() -> rocksdb::Options {
+    let mut opt = rocksdb::Options::default();
+    // Limit the total memtable memory usage per db.
+    opt.set_db_write_buffer_size(512 * 1024 * 1024);
```
huitseeker: Please use a constant at the top of the file, e.g. `DB_WRITE_BUFFER_SIZE`.

huitseeker (approved):
This and the approach you outline in [your comment](https://github.com/MystenLabs/narwhal/issues/671#issuecomment-1204782896) @mwtian, make sense to me. Thanks!

crates/typed-store/src/rocks/mod.rs
mwtian: Good idea.

PR #116, opened 2022-08-09
Allow specifying each tables options, and getting mem usage

oxade:
This PR adds:
* Two ways to configure tables at different locations
* A way to get memory usage stats, as requested by @velvia  

# Config Options:
Two ways to configure tables either at the site of struct definition, or by the opener of the DB (which overrides the former).

## 1. Base Custom Override Function 
This allows specifying a default function which is used to configure the table. 
This function is placed at the table definition, but can be overridden with approach 2

Example
```
fn custom_fn_name1() -> Options {Options::default()}
fn custom_fn_name2() -> Options {
    let mut op = custom_fn_name1();
    op.set_write_buffer_size(123456);
    op
}
#[derive(DBMapUtils)]
struct Tables {
   /// Use the functions defined
    #[defuault_options_override_fn = "custom_fn_name1"]
    table1: DBMap<String, String>,
    #[defuault_options_override_fn = "custom_fn_name2"]
    table2: DBMap<i32, String>,
    table3: DBMap<i32, String>,
    #[defuault_options_override_fn = "custom_fn_name1"]
    table4: DBMap<i32, String>,
}
```

When not specified, the default for each table is `typed_store::rocks::default_rocksdb_options()`

## 2. Caller Specified Options
This approach allows the caller of `open_tables_read_write` to specify their own table configs using a configurator.
Using this method gives fine tuned control to the caller. 

You can now supply a table `Configurator` struct to the DB before opening it

Example:
For the following struct, a TablesConfigurator struct is auto-derived with all fields defaulted to `typed_store::rocks::default_rocksdb_options()`:
```
#[derive(DBMapUtils)]
struct Tables {
    table1: DBMap<String, String>,
    table2: DBMap<i32, String>,
}
```

Get a configurator
```

    let primary_path = temp_dir();

    // Get a configurator for this table
    /* 
            The derive macro creates a new struct TablesConfigurator
            pub struct TablesConfigurator {
                pub table1 : rocksdb::Options,
                pub table2 : rocksdb::Options,
            }
    */
    
    let mut config = Tables::configurator();
```

Set the config for each table
```
    // Config table 1
    config.table1 = Options::default();
    config.table1.create_if_missing(true);
    config.table1.set_write_buffer_size(123456);

    // Config table 2
    config.table2 = config.table1.clone();

    config.table2.create_if_missing(false);
```

Open the table with the configs
```
    // Build and open with new config
    let _ = Tables::open_tables_read_write(primary_path, None, Some(config.build()));
```

# Mem Usage

Use `get_memory_usage` on Struct for mem table and cache info


crates/typed-store-macros/src/lib.rs
```diff
@@ -111,7 +111,50 @@ pub fn derive_dbmap_utils(input: TokenStream) -> TokenStream {
...
+                pub #field_names : rocksdb::Options,
+            )*
+        }
+
+        impl #config_struct_name {
+            /// Initialize to defaults
+            pub fn init() -> Self {
+                Self {
+                    #(
+                        #field_names : rocksdb::Options::default(),
```
mwtian: Can we call `default_rocksdb_options()` here instead?

oxade: we absolutely can :) I'm open to any ideas

oxade: @mwtian does the general idea make sense in the direction we're trying to go?
Or do we want more features? (particularly options per oper)

mwtian (commented):
Great! This definitely helps fine tune each rocksdb instance and column family. I wonder if we also need to provide more help for the operators of various binaries, who need blanket settings for total memory and disk space consumed by all rocksdb instances in the process. We can definitely add plumbings from user config to the options.

oxade:
> Great! This definitely helps fine tune each rocksdb instance and column family. I wonder if we also need to provide more help for the operators of various binaries, who need blanket settings for total memory and disk space consumed by all rocksdb instances in the process. We can definitely add plumbings from user config to the options.

Yes the plan is to have nicely tuned options defaults and profiles which different systems can just refer to.
Note that this config struct is optional when opening, so if not specified, a default is used, which is today's behavior.

crates/typed-store-macros/src/lib.rs
oxade: @mwtian we now call `default_rocksdb_options()`

crates/typed-store-macros-tests/src/tests/macro_tests.rs
```diff
@@ -133,3 +118,92 @@ async fn macro_test() {
     assert_eq!(format!("\"7\""), *m.get(&"\"7\"".to_string()).unwrap());
     assert_eq!(format!("\"8\""), *m.get(&"\"8\"".to_string()).unwrap());
 }
+
+/// We show that custom functions can be applied
+#[derive(DBMapUtils)]
+struct TablesCustomOptions {
+    #[defuault_options_override_fn = "another_custom_fn_name"]
```
velvia: (nit) Spelling?  `default_options_override_fn` not `defuault`

crates/typed-store-macros-tests/src/tests/macro_tests.rs
```diff
@@ -133,3 +118,92 @@ async fn macro_test() {
...
+    table1: DBMap<String, String>,
+    table2: DBMap<i32, String>,
+    #[defuault_options_override_fn = "custom_fn_name"]
+    table3: DBMap<i32, String>,
+    #[defuault_options_override_fn = "another_custom_fn_name"]
+    table4: DBMap<i32, String>,
+}
+
+lazy_static! {
```
velvia: @huitseeker and the LazyStatic police be coming for you. :)
I think we're supposed to use `OnceCell` instead

crates/typed-store-macros-tests/src/tests/macro_tests.rs
```diff
@@ -133,3 +118,92 @@ async fn macro_test() {
...
+#[derive(DBMapUtils)]
+struct TablesMemUsage {
+    table1: DBMap<String, String>,
+    table2: DBMap<i32, String>,
+    table3: DBMap<i32, String>,
+    table4: DBMap<i32, String>,
+}
+
+#[tokio::test]
+async fn macro_test_get_memory_usage() {
```
velvia, line 194: Awesome thanks so much for this!

crates/typed-store-macros/src/lib.rs
```diff
@@ -7,35 +7,41 @@ use quote::quote;
...
+/// A user defined function of signature () -> Options can be provided for each table
+/// If a an override function is not specified, the default in `typed_store::rocks::default_rocksdb_options` is used
+/// Example
+/// ```rust,ignore
+/// fn custom_fn_name() -> Options {
+///     Options::default()
+/// }
+/// ```
+/// can be used on a table by applying the attr:
+/// ``` rust,ignore
```
velvia: What does `ignore` do?  Does it ignore type checks?  We should try not to skip type checks on doc examples - they are super useful for keeping examples up to date!

crates/typed-store-macros/src/lib.rs
```diff
@@ -7,35 +7,41 @@ use quote::quote;
...
+/// #[defuault_options_override_fn = "custom_fn_name"]
```
velvia: Spelling

crates/typed-store-macros/src/lib.rs
```diff
@@ -7,35 +7,41 @@ use quote::quote;
...
+///```
 ///
 /// The typical flow for creating tables is to define a struct of DBMap tables, create the column families, then reopen
 /// If dumping is needed, there's an additional step of implementing a way to match and dump each table
```
velvia, line 36: This part of the doc is out of date right?   We now have a macro for this

crates/typed-store-macros/src/lib.rs
```diff
@@ -46,29 +52,36 @@ impl Default for TableOptions {
...
 /// Well formed struct of tables
 /// ```
+/// use rocksdb::Options;
 /// use typed_store::rocks::DBMap;
 /// use typed_store_macros::DBMapUtils;
 /// use typed_store::traits::DBMapTableUtil;
 ///
 /// /// Define a struct with all members having type DBMap<K, V>
+///
+/// fn custom_fn_name1() -> Options {Options::default()}
```
velvia, line 51: Wrap this stuff in code block and add some spaces around `Options::default()` ?

crates/typed-store-macros/src/lib.rs
```diff
@@ -218,6 +293,13 @@ pub fn derive_dbmap_utils(input: TokenStream) -> TokenStream {
                     _ => anyhow::bail!("No such table name: {}", table_name),
                 })
             }
+
+            /// This gives info about memory usage
+            fn get_memory_usage(&self) -> Result<(u64, u64), typed_store::rocks::TypedStoreError> {
```
velvia, line 290: Thanks much for this.  Maybe add `returns (total table memory, cache memory)` to the doc?

velvia (approved):
Mostly looks great @oxade !  Just some minor comments

PR #117, opened 2022-08-10
refactor: replace anyhow by eyre everywhere

huitseeker:
Better architecture of the import of `backtrace-rs`, which allows its use on stable (detail that hit us elsewhere).
Plenty of other features (colors in logs!), but mostly: unification of the libs we use across the architecture.

### DIFF
--- a/crates/component/Cargo.toml
+++ b/crates/component/Cargo.toml
@@ -7,10 +7,9 @@ description = "library for resilient component supervision"
 repository = "https://github.com/mystenlabs/mysten-infra"
 edition = "2021"
 publish = false
-
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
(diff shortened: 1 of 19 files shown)


velvia (approved):
Looks fine to me...  eyre looks like a drop in replacement.

I still have thoughts about the inefficiencies of obtaining and getting stack traces with every error, but we can debate that somewhere else.  :)

velvia:
@huitseeker was hoping you could hold off from merging this change for a few days, because @oxade has a high-pri change for me to do RocksDB memory profiling

oxade:
> @huitseeker was hoping you could hold off from merging this change for a few days, because @oxade has a high-pri change for me to do RocksDB memory profiling

@velvia I think he missed this. It's merged.
but anyways here's your [PR](https://github.com/MystenLabs/mysten-infra/pull/116) with the memory profiling work

PR #118, opened 2022-08-10
Set default max WAL size per RocksDB and allow configuration via env var

mwtian:
The goal is to mitigate the annoyance when gateway receives increased request volume, WAL logs on disk grow much larger as well.

Related to https://github.com/MystenLabs/sui/issues/3885

### DIFF
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -10,14 +10,21 @@ use bincode::Options;
 use collectable::TryExtend;
 use rocksdb::{ColumnFamilyDescriptor, DBWithThreadMode, MultiThreaded, WriteBatch};
 use serde::{de::DeserializeOwned, Serialize};
-use std::{borrow::Borrow, marker::PhantomData, path::Path, sync::Arc};
-use tracing::instrument;
+use std::{borrow::Borrow, env, marker::PhantomData, path::Path, sync::Arc};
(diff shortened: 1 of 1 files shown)


crates/typed-store/src/rocks/mod.rs
```diff
@@ -430,11 +437,46 @@ where
...
+                    var_name, e
+                );
+                Option::None
+            }
+        },
+        Err(e) => {
+            info!("Env var {} is not set: {}", var_name, e);
+            Option::None
+        }
+    }
```
huitseeker, line 456: 1. `Option::None` is pedantic, `None` would suffice.
2. Consider using `and_then`:
`env::var(var_name).and_then(|v| v.parse::<usize>()).ok()`
3. and then to implement your rich logging, you can use [tap](https://docs.rs/tap/latest/tap/index.html), specifically `tap_err`.

huitseeker (approved):
LGTM, I left stylistic comments.

crates/typed-store/src/rocks/mod.rs
mwtian: Sorry I turned on auto merge. Will address the feedback in another PR.

huitseeker: No worries, those are stylistic recommendations, and aside from the first, not really binding.

PR #119, opened 2022-08-13
Cleanup: use one expression to read usize from env var

mwtian:
This addresses the comment at https://github.com/MystenLabs/mysten-infra/pull/118#discussion_r943921496

### DIFF
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -11,6 +11,7 @@ use collectable::TryExtend;
 use rocksdb::{ColumnFamilyDescriptor, DBWithThreadMode, MultiThreaded, WriteBatch};
 use serde::{de::DeserializeOwned, Serialize};
 use std::{borrow::Borrow, env, marker::PhantomData, path::Path, sync::Arc};
+use tap::TapFallible;
 use tracing::{info, instrument};
 
 use self::{iter::Iter, keys::Keys, values::Values};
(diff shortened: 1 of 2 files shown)


crates/typed-store/src/rocks/mod.rs
```diff
@@ -438,22 +439,17 @@ where
...
+        .tap_err(|e| info!("Env var {} is not set: {}", var_name, e))
+        .ok()?
+        .parse::<usize>()
+        .tap_err(|e| {
+            info!(
+                "Env var {} does not contain valid usize integer: {}",
+                var_name, e
+            )
+        })
+        .ok()
```
huitseeker, line 452:
```suggestion
    env::var(var_name).and_then(|v|.parse::<usize>()
        .tap_err(|e| {
            info!(
                "Env var {} unset or does not contain valid usize integer: {}",
                var_name, e
            )
        }))
        .ok()
```

huitseeker (approved):
Thanks!

crates/typed-store/src/rocks/mod.rs
mwtian: It seems the signature of `Result<U, E>::and_then()` is 
```
pub fn and_then<U, F: FnOnce(T) -> Result<U, E>>(self, op: F) -> Result<U, E>
```
So the `Result` and the closure in `and_then()` must have the same error type. Here `env::var()` and `parse::<usize>()` have different error types.

huitseeker: Yup, I'd expect a `map_err` to come in at some point. I trust you to find the best compromise.

mwtian: Sounds good.

PR #120, opened 2022-08-13
add bytes to trace layer

lanvidr:
https://github.com/MystenLabs/sui/issues/3835


huitseeker (changes requested):
Thanks for pushing this!
I worry that this sort of logging is extremely prone to pushing very large amounts of logs that may be spammy. Wouldn't this be better served by a metric?

lanvidr:
Metrics would be my preference as well, and are in progress, although the metrics will be rough estimations. It was mentioned to have logs in addition to the metrics [here]. (https://github.com/MystenLabs/sui/issues/3882)

PR #123, opened 2022-08-18
utils for Store open and config

### LINKED ISSUE #671: [NodeStorage] Expose the ability to set custom rocksdb::Options
Currently we open the NodeStorage database with None as options, which does not allow callers (like Sui) to set limits on the size of memtables. We should allow opening with custom DB Options:
https://github.com/MystenLabs/narwhal/blob/442d853cf92bc974fc66420b5c77ea6986db3bbb/node/src/lib.rs#L57-L70

oxade:
Basically implements some `DBMapUtil` features for `Store`.
We will unify them in subsequent PRs

TODO: extend this PR for dumping `Store`

This also adds the `configurator` which allows us to close https://github.com/MystenLabs/narwhal/issues/671


crates/typed-store-macros/src/lib.rs
```diff
@@ -409,3 +410,217 @@ fn extract_generics_names(generics: &Generics) -> Vec<Ident> {
...
+        "core::hash::Hash + core::cmp::Eq + std::fmt::Debug + serde::Serialize + for<'de> serde::de::Deserialize<'de> + core::marker::Send + 'static";
+    let generics_bounds_token: proc_macro2::TokenStream = generics_bounds.parse().unwrap();
+
+    let config_struct_name_str = format!("{}Configurator", name);
+    let config_struct_name: proc_macro2::TokenStream = config_struct_name_str.parse().unwrap();
+
+    // let first_field_name = field_names
+    //     .get(0)
+    //     .expect("Expected at least one field")
+    //     .clone();
```
patrickkuo: unwanted?

crates/typed-store-macros/src/lib.rs
```diff
@@ -409,3 +410,217 @@ fn extract_generics_names(generics: &Generics) -> Vec<Ident> {
...
+
+    let _simple_field_type_name: proc_macro2::TokenStream = simple_field_type_name.parse().unwrap();
```
patrickkuo: do we need this?

patrickkuo (approved):
2 comments on unwanted code, other LGTM

crates/typed-store-macros/src/lib.rs
```diff
@@ -409,3 +410,217 @@ fn extract_generics_names(generics: &Generics) -> Vec<Ident> {
...
+/// async fn main() {
+/// let primary_path = tempfile::tempdir().expect("Failed to open temporary directory").into_path();
+/// /// This is auto derived
+/// let tbls_primary = Tables::open_tables_read_write(primary_path.clone(), None, None);
+///
+/// // Do some stuff with the DB
+/// }
+///
+/// // Bad usage example
+/// // Structs fields most only be of type Store<K, V>
```
huitseeker: Typo

velvia:
Hey @oxade .... I think we should discuss on this PR and hold off a little bit.

I'm not sure I understand the need to rename `DBMap` to `Store`....  before you had a macro that derived methods for each table and we did not need to rename DBMap.  Why not keep that?

I know there are concerns from others about having too much magic, and if we should have less.   I think changing the name from `DBMap` to `Store` might be going in the wrong direction, so that discussion needs to be resolved.

Personally I don't think `Store` is a good name.... `DBMap` is pretty precise, as it is a Map-like API which is part of a database.  However, `Store` is very generic and usually when I think of `Store` it is like `TransactionStore`, `EventStore`, `IndexStore` etc...  a store is composed of multiple tables or maps, not a single map.     There could be a better name but I'm not sure we can improve upon `DBMap` for the current use case.

Perhaps I don't understand the long term vision and we need to lay that out first and discuss it..... thanks

oxade:
> Hey @oxade .... I think we should discuss on this PR and hold off a little bit.
> 
> I'm not sure I understand the need to rename `DBMap` to `Store`.... before you had a macro that derived methods for each table and we did not need to rename DBMap. Why not keep that?
> 
> I know there are concerns from others about having too much magic, and if we should have less. I think changing the name from `DBMap` to `Store` might be going in the wrong direction, so that discussion needs to be resolved.
> 
> Personally I don't think `Store` is a good name.... `DBMap` is pretty precise, as it is a Map-like API which is part of a database. However, `Store` is very generic and usually when I think of `Store` it is like `TransactionStore`, `EventStore`, `IndexStore` etc... a store is composed of multiple tables or maps, not a single map. There could be a better name but I'm not sure we can improve upon `DBMap` for the current use case.
> 
> Perhaps I don't understand the long term vision and we need to lay that out first and discuss it..... thanks

@velvia I think you misunderstand
We're not renaming.
`Store` has always[ existed](https://github.com/MystenLabs/mysten-infra/blob/main/crates/typed-store/src/lib.rs#L50). It is the async counterpart to our normal tables and used DBMap underneath. It's just a wrapper with some channels and is used by Narwhal. This PR does not affect Sui

But yes the original name `Store` should probably change to be more clear.

crates/typed-store-macros-tests/src/tests/macro_tests.rs
```diff
@@ -204,3 +205,58 @@ async fn macro_test_get_memory_usage() {
     let (mem_table, _) = tables.get_memory_usage().unwrap();
     assert!(mem_table > 0);
 }
+
+#[derive(StoreUtils)]
+struct StoreTables {
+    table1: Store<Vec<u8>, Vec<u8>>,
+    _table2: Store<i32, String>,
```
huitseeker: Nit: the underscore isn't necessary I think.

crates/typed-store-macros/src/lib.rs
```diff
@@ -409,3 +410,217 @@ fn extract_generics_names(generics: &Generics) -> Vec<Ident> {
...
+        .iter()
+        .map(|q| {
+            let GeneralTableOptions::OverrideFunction(fn_name) = q;
+            fn_name.parse().unwrap()
+        })
+        .collect();
+
+    let generics_bounds =
+        "core::hash::Hash + core::cmp::Eq + std::fmt::Debug + serde::Serialize + for<'de> serde::de::Deserialize<'de> + core::marker::Send + 'static";
+    let generics_bounds_token: proc_macro2::TokenStream = generics_bounds.parse().unwrap();
```
huitseeker: You can use `syn::parse_quote` to make this much nicer:
```rust
fn add_basic_bound(mut generics: Generics) -> Generics {
    for param in &mut generics.params {
        if let GenericParam::Type(ref mut type_param) = *param {
            type_param
                .bounds
                .push(parse_quote!(::my::generic::Bounds));
        }
    }
    generics
}
```

crates/typed-store-macros/src/lib.rs
```diff
@@ -409,3 +410,217 @@ fn extract_generics_names(generics: &Generics) -> Vec<Ident> {
...
+
+    let config_struct_name_str = format!("{}Configurator", name);
+    let config_struct_name: proc_macro2::TokenStream = config_struct_name_str.parse().unwrap();
```
huitseeker: By and large, every time you use `parse()`, ask yourself if you can reformulate with [parse_quote](https://docs.rs/syn/latest/syn/macro.parse_quote.html)

crates/typed-store-macros/src/lib.rs
```diff
@@ -409,3 +410,217 @@ fn extract_generics_names(generics: &Generics) -> Vec<Ident> {
...
+            /// Only one process is allowed to do this at a time
+            /// `global_db_options_override` apply to the whole DB
+            /// `tables_db_options_override` apply to each table. If `None`, the attributes from `default_options_override_fn` are used if any
+            fn open_tables_read_write(
+                path: std::path::PathBuf,
+                global_db_options_override: Option<rocksdb::Options>,
+                tables_db_options_override: Option<typed_store::rocks::DBMapTableConfigMap>
+            ) -> Self {
+                Self::open_tables_impl(path, None, global_db_options_override, tables_db_options_override)
+            }
```
huitseeker: This strongly hints at the idea that `StoreTableUtil` could be a trait with one required method (`open _tables_impl`), and a few methods with default implementations, such as this one. Then, the macro would only have to generate the required method in the impl.

huitseeker (commented):
Thanks for the PR!

oxade:
Cleanups with parse_quote will be done here https://github.com/MystenLabs/sui/issues/6012

PR #125, opened 2022-08-21
[typed-store] implement the prev() on iterator

akichidis:
Introduces the method `prev()` on the iterator used by the `DBMap` so we can retrieve our results in reverse order. That's particularly useful in situations where you want to `skip_to_last` element and start retrieving the elements backwards.

### DIFF
--- a/crates/typed-store/src/rocks/iter.rs
+++ b/crates/typed-store/src/rocks/iter.rs
@@ -3,6 +3,7 @@
 use std::marker::PhantomData;
 
 use bincode::Options;
+use rocksdb::Direction;
 
 use super::{be_fix_int_ser, errors::TypedStoreError};
 use serde::{de::DeserializeOwned, Serialize};
@@ -13,6 +14,7 @@ use super::DBRawIteratorMultiThreaded;
 pub struct Iter<'a, K, V> {
     db_iter: DBRawIteratorMultiThreaded<'a>,
     _phantom: PhantomData<(K, V)>,
+    direction: Direction,
 }
 
 impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iter<'a, K, V> {
@@ -20,24 +22,7 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iter<'a, K, V> {
         Self {
             db_iter,
             _phantom: PhantomData,
-        }
-    }
-
-    pub fn prev(&mut self) -> Option<(K, V)> {
-        if self.db_iter.valid() {
-            let config = bincode::DefaultOptions::new()
-                .with_big_endian()
-                .with_fixint_encoding();
-            let key = self.db_iter.key().and_then(|k| config.deserialize(k).ok());
-            let value = self
-                .db_iter
-                .value()
-                .and_then(|v| bincode::deserialize(v).ok());
-
-            self.db_iter.prev();
-            key.and_then(|k| value.map(|v| (k, v)))
-        } else {
-            None
+            direction: Direction::Forward,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


huitseeker (commented):
I see no objection with this but I note that the idiomatic way to express this is to [implement a `DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) rather than add a method on the `Iter` implementation which would be difficult to discover.
What do you think about this approach?

akichidis:
> I see no objection with this but I note that the idiomatic way to express this is to [implement a `DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) rather than add a method on the `Iter` implementation which would be difficult to discover. What do you think about this approach?

Ah nice, thanks @huitseeker for pointing me on the DoubleEndedIterator, wasn't aware of it, makes total sense! I am updating the PR now

huitseeker (commented):
I've added a test to the end of this PR, illustrating why I think the behavior of the `DoubleEndedIterator` doesn't match its specification - the implementation iterates in the reverse direction, but is supposed to remove from the end of the iterator.
This may require modifying the `Iter` struct to keep two pointers to iterators with distinct [`IteratorMode`](https://docs.rs/rocksdb/latest/rocksdb/enum.IteratorMode.html) values.

(should you not want to get into it at this time, tabling this and going back to an implementation of `prev` is fine too)

crates/typed-store/src/rocks/tests.rs
```diff
@@ -233,6 +233,26 @@ fn test_iter() {
...
+
+    db.insert(&1, &"1".to_string()).expect("Failed to insert");
+    db.insert(&2, &"2".to_string()).expect("Failed to insert");
+    db.insert(&3, &"3".to_string()).expect("Failed to insert");
+
+    let mut iter = db.iter().skip_to_last().reverse();
+    assert_eq!(Some((3, "3".to_string())), iter.next());
+    assert_eq!(Some((2, "2".to_string())), iter.next());
+    assert_eq!(Some((1, "1".to_string())), iter.next());
+    assert_eq!(None, iter.next());
```
akichidis, line 248: I believe this is now more _natural_ when it comes to iterate the elements in reverse order. Thoughts @huitseeker ?

crates/typed-store/src/rocks/iter.rs
```diff
@@ -68,4 +75,35 @@ impl<'a, K: Serialize, V> Iter<'a, K, V> {
...
+        self.direction = Direction::Reverse;
+        RevIter::new(self)
+    }
+}
+
+/// An iterator with a reverted direction to the original. The `RevIter`
+/// is hosting an iteration which is consuming in the opposing direction.
+/// It's not possible to do further manipulation (ex re-reverse) to the
+/// iterator.
+pub struct RevIter<'a, K, V> {
```
akichidis, line 92: Introduced a separate struct so can limit the user from doing any complex combinations of seeks that could confuse the iterator. Now there is only one option once you reverse...you are calling `next` to traverse the results backwards

crates/typed-store/src/rocks/tests.rs
huitseeker: 👍  Yep. Perhaps this departs slightly from the rust idiom, but this is definitely correct, which is what I think we care about.

huitseeker (approved):
Thanks a bunch @akichidis 🙏

crates/typed-store/src/rocks/tests.rs
akichidis: Ah yeah it's not very Rusty , but tried to go for correctness here as you said

PR #126, opened 2022-08-22
Expose RocksDB transactions in mysten infra

sadhansood:
This PR adds the ability to perform optimistic transactions on RocksDB in mysten wrapper lib. We do several workarounds  to overcome this limitation in Sui(maybe Narwhal too?) which are sub-optimal and limits performance severly. OptimisticTransactionDB has couple limitation which are not very hard (seemingly) to fix:
1. TxDB does not have support for getting memory stats but looking at rust-rocksdb it should be possible to submit a PR to add that feature.
2. Delete range is unsupported in transaction db but doing a code search, we don't use it anywhere anyways.

### DIFF
--- a/crates/typed-store-macros-tests/src/tests/macro_tests.rs
+++ b/crates/typed-store-macros-tests/src/tests/macro_tests.rs
@@ -201,6 +201,6 @@ async fn macro_test_get_memory_usage() {
         .multi_insert(keys_vals_1)
         .expect("Failed to multi-insert");
 
-    let (mem_table, _) = tables.get_memory_usage().unwrap();
-    assert!(mem_table > 0);
+    // let (mem_table, _) = tables.get_memory_usage().unwrap();
+    // assert!(mem_table > 0);
 }
--- a/crates/typed-store-macros/src/lib.rs
+++ b/crates/typed-store-macros/src/lib.rs
@@ -227,7 +227,7 @@ pub fn derive_dbmap_utils(input: TokenStream) -> TokenStream {
 
                     let res = match with_secondary_path {
                         Some(p) => typed_store::rocks::open_cf_opts_secondary(path, Some(&p), global_db_options_override, &opt_cfs),
-                        None    => typed_store::rocks::open_cf_opts(path, global_db_options_override, &opt_cfs)
+                        None    => typed_store::rocks::open_cf_opts_transaction_db(path, global_db_options_override, &opt_cfs)
                     };
                     res
                 }.expect("Cannot open DB.");
@@ -288,8 +288,7 @@ pub fn derive_dbmap_utils(input: TokenStream) -> TokenStream {
 
             /// This gives info about memory usage and returns a tuple of total table memory usage and cache memory usage
             fn get_memory_usage(&self) -> Result<(u64, u64), typed_store::rocks::TypedStoreError> {
-                let stats = rocksdb::perf::get_memory_usage_stats(Some(&[&self.#first_field_name.rocksdb]), None)
-                    .map_err(|e| typed_store::rocks::TypedStoreError::RocksDBError(e.to_string()))?;
+                let stats = self.#first_field_name.rocksdb.get_memory_usage_stats()?;
                 Ok((stats.mem_table_total, stats.cache_total))
             }
         }
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -8,14 +8,19 @@ mod values;
 use crate::traits::Map;
 use bincode::Options;
 use collectable::TryExtend;
-use rocksdb::{ColumnFamilyDescriptor, DBWithThreadMode, MultiThreaded, WriteBatch};
+use rocksdb::{
+    perf::MemoryUsageStats, AsColumnFamilyRef, BoundColumnFamily, ColumnFamilyDescriptor,
+    DBPinnableSlice, DBWithThreadMode, Error, MultiThreaded, OptimisticTransactionDB, Transaction,
+    WriteBatchWithTransaction,
+};
 use serde::{de::DeserializeOwned, Serialize};
 use std::{borrow::Borrow, collections::BTreeMap, env, marker::PhantomData, path::Path, sync::Arc};
 use tap::TapFallible;
 use tracing::{info, instrument};
 
-use self::{iter::Iter, keys::Keys, values::Values};
 pub use errors::TypedStoreError;
+
+use self::{iter::Iter, keys::Keys, values::Values};
 
 // Write buffer size per RocksDB instance can be set via the env var below.
 // If the env var is not set, use the default value in MiB.
@@ -29,9 +34,6 @@ const DEFAULT_DB_WAL_SIZE: usize = 1024;
 
 #[cfg(test)]
 mod tests;
-
-type DBRawIteratorMultiThreaded<'a> =
-    rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>;
 
 /// A helper macro to reopen multiple column families. The macro returns
 /// a tuple of DBMap structs in the same order that the column families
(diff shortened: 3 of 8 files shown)


crates/typed-store-macros-tests/src/tests/macro_tests.rs
```diff
@@ -201,6 +201,6 @@ async fn macro_test_get_memory_usage() {
         .multi_insert(keys_vals_1)
         .expect("Failed to multi-insert");
 
-    let (mem_table, _) = tables.get_memory_usage().unwrap();
```
oxade, line 204: Why remove this? 
And it's not relevant, delete instead of comment 
The original intent was to make sure the function indeed is callable.

crates/typed-store-macros/src/lib.rs
```diff
@@ -288,8 +288,7 @@ pub fn derive_dbmap_utils(input: TokenStream) -> TokenStream {
 
             /// This gives info about memory usage and returns a tuple of total table memory usage and cache memory usage
             fn get_memory_usage(&self) -> Result<(u64, u64), typed_store::rocks::TypedStoreError> {
-                let stats = rocksdb::perf::get_memory_usage_stats(Some(&[&self.#first_field_name.rocksdb]), None)
-                    .map_err(|e| typed_store::rocks::TypedStoreError::RocksDBError(e.to_string()))?;
+                let stats = self.#first_field_name.rocksdb.get_memory_usage_stats()?;
                 Ok((stats.mem_table_total, stats.cache_total))
             }
```
huitseeker, line 293: Please check we do not generate this in a breaking fashion (which seems hinted at by its commented nature in the macro tests). If we can't implement it for now, I'd rather not have this at all and open an issue to contribute it upstream.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -74,10 +76,228 @@ macro_rules! reopen {
...
+        }
+    }
+}
+
+pub enum RocksDBIter<'a> {
+    DB(rocksdb::DBRawIteratorWithThreadMode<'a, DBWithThreadMode<MultiThreaded>>),
+    OptimisticTransactionDB(
+        rocksdb::DBRawIteratorWithThreadMode<'a, OptimisticTransactionDB<MultiThreaded>>,
+    ),
+    OptimisticTransaction(
```
huitseeker, line 223: I'm confused: while it's clear _how_ you're developing this capability, what is the objective of iterating over a transaction today, in this PR?
Does that objective require defining iteration over this transaction alongside, and a equivalent to, iterating over a DB (optimistic-transactional or otherwise)?

crates/typed-store/src/rocks/tests.rs
```diff
@@ -540,3 +549,103 @@ async fn open_as_secondary_test() {
...
+        // modify same key in another transaction
+        let txn2 = db_cf_1.optimistic_transaction().unwrap();
+        assert!(txn2
+            .insert(&db_cf_1, &123456789, &"66666".to_string())
+            .is_ok());
+        assert!(txn2.commit().is_ok());
+
+        // txn1 should fail with RocksDBError
+        let err = txn1.commit().unwrap_err();
+        assert!(matches!(err, TypedStoreError::RocksDBError { .. }));
```
huitseeker, line 601: I'd be more precise in the match, which IIUC should be sth like:
```
assert_eq!(match err {
    TypedStoreError::RocksDBError { e } => Some(e.kind()),
    _ => None,
}, Some(ErrorKind::Busy))
```

crates/typed-store/src/rocks/mod.rs
```diff
@@ -74,10 +76,228 @@ macro_rules! reopen {
...
+            RocksDB::OptimisticTransactionDB(_) => RocksDBWriteBatch::OptimisticTransactionDB(
+                WriteBatchWithTransaction::<true>::default(),
+            ),
+        }
+    }
+    pub fn get_memory_usage_stats(&self) -> Result<MemoryUsageStats, TypedStoreError> {
+        match self {
+            RocksDB::DB(db) => rocksdb::perf::get_memory_usage_stats(Some(&[db]), None)
+                .map_err(|e| TypedStoreError::RocksDBError(e.into())),
+            RocksDB::OptimisticTransactionDB(_db) => {
```
huitseeker, line 211: I wonder if this is the appropriate error here?

huitseeker (changes requested):
Thanks a lot for this PR @sadhansood ! We've been waiting for this capability nearly since the company was started, and since the upstream PR has finally been merged, we can finally have some of it! This is an awesome direction, and I'm very grateful for the PR.

I have a few areas of concern though. 1 & 2 are really design questions, 3 marks a large loss in test coverage (and is the main reason why this review is requesting changes).

1. **Isolation:** We should systematically point the user of *any* transaction features to https://github.com/facebook/rocksdb/wiki/Transactions#optimistictransactiondb 
I do not expect everyone to be familiar with optimistic concurrency control, and I do not think everyone will be able to interpret its unhappy path without surprise.
In particular, we do not set snapshots, or open an option to do so here (perhaps we should), and I expect this isolation contract to trip up users if not spectacularly highlighted:
    > By default, Transaction conflict checking validates that no one else has written a key _after_ the time the key was first written in this transaction. This isolation guarantee is sufficient for many use-cases. However, you may want to guarantee that no else has written a key since the start of the transaction. This can be accomplished by calling SetSnapshot() after creating the transaction.
    
    The example that immediately follows this sentence in the doc is striking. As an aside, I remark the PR does not expose [`get_for_update` either](https://github.com/rust-rocksdb/rust-rocksdb/blob/9118a60fe265f4c425a162a3958a96e008be646f/tests/test_optimistic_transaction_db.rs#L413-L424).
    
    Overall, this is problematic for the use case of using transactionality in the  `authority_store` to manage locks: there, we want a multi-key CAS: the guarantee that no other task has touched any of a set of locks between reading all of them (to check all objects are unlocked) and writing all of them transactionally. It seems to me that at least there, we want strong isolation.

2. **Future steps / Use case doc:** I wonder about the following warning found in the doc:
    > However, this may not be a good solution for workloads where write-conflicts occur frequently due to many transactions constantly attempting to update the same keys. For these workloads, using a TransactionDB may be a better fit. 

    Should we consider *three* variants in the `RocksDB` enum (normal, transactional, optimistic-transactional)? Should we, more prosaically, just expose `Transaction` (and use it just in the `authority _store` and maybe checkpoints)? What performance should we expect there?
    I expect `OptimisticTransactionDB` to be an easier-to-configure option, at the cost of significant investment from developers in the unhappy path: could we better equip them to deal with this?
    Would a handful of scenario-driven code examples help with this?
    
3. **Test regression:** This PR changes the behavior of the tests to only test the `OptimisticTransactionDB` variant of the `RocksDB` enum. The `DB` variant is left untested. Yet, the tests for the `DB` variant will continue to remain relevant for backwards-compatibility and db usage that does not require transactional behavior (e.g. Narwhal).
The code path that creates this situation is `DBMap::open` -> `DBMap::open_cf` -> `DBMap::open_cf_opts_transaction_db`

4. **Addendum:** You mention:
    > Delete range is unsupported in transaction db but doing a code search, we don't use it anywhere anyways.
    
    We do!
    ```
    huitseeker@dapple.localdomain➜~/tmp/sui(rocksdb_warts)» rg multi_remove
    crates/sui-storage/src/lock_service.rs
    323:        self.transaction_lock.multi_remove(objects)?;
    ```
    It may well be that this is inconsequential, but this does seem to warrant further exploration of the specific locks use case (perhaps through an [example](http://xion.io/post/code/rust-examples.html)). I would also recommend @mystenmark and @gdanezis to have a look at this super-interesting PR!

oxade (changes requested):

mystenmark:
@sadhansood Thanks for doing this, there have definitely been times I've wished to be able to do this. Some comments/questions:

1. Can you explain which work-arounds you have in mind in Sui, and how you've determined how they limit performance? In my mind, the main enabler of performance in Sui is that the design does not require very many consistency guarantees in principle, so in fact we won't need transactions nor work-arounds for the lack thereof in the majority of cases.

   The main exception is the lock database. Right now we overcome the lack of optimistic transactions there by making the lock service single threaded. Clearly that will limit performance at some point, but my question is if that is a bottleneck right now? And if so how severe? There are other options we could use to make the lockservice multithreaded, e.g. using a mutex table to prevent concurrent object read/writes.

   I don't have a personal preference for one solution over the other, but if the motivation of this PR is performance, then I think we need to have some measurements (preferably in the form of benchmarks) to guide us here.

1. It would be very helpful to see example code showing how the new APIs would be used in practice. Re-implementing https://github.com/MystenLabs/sui/blob/main/crates/sui-storage/src/lock_service.rs#L233 as an example would be very illustrative.

1. I'm unclear (to me) from the diffs what isolation level this provides. In my opinion, it is critical that we expose serializable isolation by default - any weaker form will almost surely be misused. If we need weaker forms for performance in certain cases, that should be taken on a case-by-case basis.

PR #131, opened 2022-08-25
[Feature][Telemetry] Automatic span latency measurement

velvia:
This PR adds a custom `tracing` `Layer` that records the latency of every `Span` into a Prometheus Histogram.
A label `span_name` contains the name of the span, so one can easily search for latency stats for a particular span they are interested in.

The memory usage of this layer is one Prometheus Histogram (with a configurable number of buckets, but at least 10 is recommended) per declared span in the codebase.

This PR should make it vastly easier to track the running latency statistics of any tracing span declared in an application.

### DIFF
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -17,6 +17,7 @@
 //! - `json` - Bunyan formatter - JSON log output, optional
 //! - `tokio-console` - [Tokio-console](https://github.com/tokio-rs/console) subscriber, optional
 
+use span_latency_prom::PrometheusSpanLatencyLayer;
 use std::{
     env,
     io::{stderr, Write},
@@ -46,7 +47,7 @@ pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 /// - log_level: error/warn/info/debug/trace, defaults to info
 /// - service_name:
 #[derive(Default, Clone, Debug)]
-pub struct TelemetryConfig {
+pub struct TelemetryConfig<'reg> {
     /// The name of the service for Jaeger and Bunyan
     pub service_name: String,
 
@@ -65,6 +66,8 @@ pub struct TelemetryConfig {
     pub panic_hook: bool,
     /// Crash on panic
     pub crash_on_panic: bool,
+    /// Optional Prometheus registry - if present, all enabled span latencies are measured
+    pub prom_registry: Option<&'reg prometheus::Registry>,
 }
 
 #[must_use]
@@ -142,7 +145,7 @@ fn set_panic_hook(crash_on_panic: bool) {
     }));
 }
 
-impl TelemetryConfig {
+impl<'reg> TelemetryConfig<'reg> {
     pub fn new(service_name: &str) -> Self {
         Self {
             service_name: service_name.to_owned(),
@@ -154,6 +157,7 @@ impl TelemetryConfig {
             log_string: None,
             panic_hook: true,
             crash_on_panic: false,
+            prom_registry: None,
         }
     }
 
@@ -164,6 +168,11 @@ impl TelemetryConfig {
 
     pub fn with_log_file(mut self, filename: &str) -> Self {
         self.log_file = Some(filename.to_owned());
+        self
+    }
+
+    pub fn with_prom_registry(mut self, registry: &'reg prometheus::Registry) -> Self {
+        self.prom_registry = Some(registry);
         self
     }
 
@@ -221,6 +230,12 @@ impl TelemetryConfig {
         } else {
             None
         };
+
+        if let Some(registry) = config.prom_registry {
+            let span_lat_layer = PrometheusSpanLatencyLayer::try_new(registry, 15)
+                .expect("Could not initialize span latency layer");
+            layers.push(span_lat_layer.boxed());
+        }
 
         #[cfg(feature = "jaeger")]
         if config.enable_tracing {
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -34,12 +34,14 @@ use tracing_subscriber::{
 
 use crossterm::tty::IsTty;
 
+pub mod span_latency_prom;
+
 /// Alias for a type-erased error type.
 pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
 /// Configuration for different logging/tracing options
 /// ===
-/// - json_log_output: Output JSON logs to stdout only.  No other options will work.
+/// - json_log_output: Output JSON logs to stdout only.
 /// - log_file: If defined, write output to a file starting with this name, ex app.log
 /// - log_level: error/warn/info/debug/trace, defaults to info
 /// - service_name:
@@ -49,8 +51,9 @@ pub struct TelemetryConfig {
     pub service_name: String,
 
     pub enable_tracing: bool,
+    /// Enables Tokio Console debugging on port 6669
     pub tokio_console: bool,
-    /// Output JSON logs.  Tracing and Tokio Console are not available if this is enabled.
+    /// Output JSON logs.
     pub json_log_output: bool,
     /// Write chrome trace output, which can be loaded from chrome://tracing
     pub chrome_trace_output: bool,
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -354,8 +354,6 @@ mod tests {
         assert_eq!(labels[0].get_value(), "yo span yo");
 
         panic!("This should cause error logs to be printed out!");
-
-        // Now, latencies from the span should be visible in the registry
     }
 
     /*
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -323,21 +323,45 @@ pub fn init_for_testing() {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use tracing::{debug, info, warn};
+    use prometheus::proto::MetricType;
+    use std::time::Duration;
+    use tracing::{debug, info, info_span, warn};
 
     #[test]
     #[should_panic]
     fn test_telemetry_init() {
-        let _guard = TelemetryConfig::new("my_app").init();
+        let registry = prometheus::Registry::new();
+        let config = TelemetryConfig::new("my_app").with_prom_registry(&registry);
+        let _guard = config.init();
 
         info!(a = 1, "This will be INFO.");
-        debug!(a = 2, "This will be DEBUG.");
-        warn!(a = 3, "This will be WARNING.");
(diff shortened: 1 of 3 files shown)


velvia:
Need some help with cargo-deny failure:
```
error[A001]: Potential segfault in the time crate
[14](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:15)
    ┌─ /github/workspace/Cargo.lock:268:1
[15](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:16)
    │
[16](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:17)
268 │ time 0.1.44 registry+https://github.com/rust-lang/crates.io-index
[17](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:18)
    │ ----------------------------------------------------------------- security vulnerability detected
[18](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:19)
    │
[19](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:20)
    = ID: RUSTSEC-2020-0071
[20](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:21)
    = Advisory: https://rustsec.org/advisories/RUSTSEC-2020-0071
[21](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:22)
    = ### Impact
[22](https://github.com/MystenLabs/mysten-infra/runs/8007709614?check_suite_focus=true#step:4:23)
```

It seems we have two versions of time in dependencies, and one of them comes from libc... which is weird.

```
  time@0.1.44
  time@0.3.11
~/s/m/mysten-infra (ec/span-latency-measurement ↩=) cargo tree -p time@0.1.44
time v0.1.44
└── libc v0.2.126
```

crates/telemetry-subscribers/src/span_latency_prom.rs
```diff
@@ -0,0 +1,102 @@
...
+        id: &span::Id,
+        ctx: tracing_subscriber::layer::Context<S>,
+    ) {
+        let span = ctx.span(id).unwrap();
+        // NOTE: there are other extensions that insert timings.  For example,
+        // tracing_subscriber's with_span_events() inserts events at open and close that contain timings.
+        // However, we cannot be guaranteed that those events would be turned on.
+        // TODO: maybe add a check if something else is already inserted...
+        span.extensions_mut()
+            .insert(PromSpanTimestamp(Utc::now().timestamp()));
```
andll: `Instant::now` rather then `Utc` is better fit for local time measurement 
- it is way faster: Instance::now is a register read, while Utc::now is often a syscall
- it is also more reliable: local clock returned by Utc can go back and forth while Instant is monotonic

crates/telemetry-subscribers/src/span_latency_prom.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// uses up in Prometheus (and in the application).  10 is probably a minimum.
+    pub fn try_new(registry: &Registry, num_buckets: usize) -> Result<Self, PrometheusSpanError> {
+        if num_buckets < 1 {
+            return Err(PrometheusSpanError::ZeroOrNegativeNumBuckets);
+        }
+
+        // Histogram for span latencies must accommodate a wide range of possible latencies, so
+        // don't use the default Prometheus buckets
+        // The latencies are in nanoseconds
+        let factor = (TOP_LATENCY_IN_NS / LOWEST_LATENCY_IN_NS).powf(1.0 / (num_buckets as f64));
```
longbowlu, line 57: any reference on how this formula?

crates/telemetry-subscribers/src/span_latency_prom.rs
```diff
@@ -0,0 +1,102 @@
...
+        &self,
+        _attrs: &span::Attributes,
+        id: &span::Id,
+        ctx: tracing_subscriber::layer::Context<S>,
+    ) {
+        let span = ctx.span(id).unwrap();
+        // NOTE: there are other extensions that insert timings.  For example,
+        // tracing_subscriber's with_span_events() inserts events at open and close that contain timings.
+        // However, we cannot be guaranteed that those events would be turned on.
+        // TODO: maybe add a check if something else is already inserted...
```
longbowlu: do we care? they are not `PromSpanTimestamp`

longbowlu (commented):
overall lgtm. 

can we optionally toggle on/off metrics collections from some spans ? I'm afraid having this automatically enabled for every span will accidentally consume more resources then we want in the future.

crates/telemetry-subscribers/src/span_latency_prom.rs
velvia: Ok, sounds good

crates/telemetry-subscribers/src/span_latency_prom.rs
velvia: It's pretty simple, you are just calculating the multiplier that can get you from LOWEST to TOP in `num_buckets` steps, which is basically the nth root of the (TOP/LOWEST) ratio.   I can add a comment

crates/telemetry-subscribers/src/span_latency_prom.rs
velvia: Well, I was originally hoping that somebody else would insert timings and we could just use it.  There is nothing that is guaranteed, but there is something if spans are logged.

velvia:
@longbowlu thanks for the look.  I understand the concern about using too many resources, but I don't think this is something to worry about.   The reasons:

1. The biggest consumer of resources (memory, locking) is the creation of a span in the first place, as it adds a lot of metadata for each span.  There are multiple spans per transaction, so each time a transaction is run, maybe 5 debug spans are created today in validator (and maybe 5 more in quorum_driver).
2. However, this PR only uses resources per type of span (to be more exact, per span name).    It uses a Prometheus histogram. It also adds a timing, which is very small compared to the rest of the span metadata.

Today as I said there are something like 5 debug spans and 10-15 trace spans in the validator, and some more in the quorum_driver.  We will add more too.   By default, trace logging is off, so we only see debug.

If we log span creation and span close, like we do today, then regardless of whether this PR's histogram feature is turned on or off, we are using a ton of resources in creating spans.

Turning on this feature will create, by default, 5 Prometheus latency histograms, for the lifetime of the node.   That's it.    And a tiny amount (basically a u64) of additional state per span.   The histogram memory usage can be further tuned by tuning the number of buckets, but again it is trivial compared to the memory used per span created (used for logging).

We have further mechanisms to control span creation in the future:
1. First, logging levels dictate which spans are enabled or not, today
2. I have already filed an issue for span sampling.  This would be the answer in the future to the resource usage of the spans - that is we should not create spans every time, but only a fraction of the time, and it should be done intelligently so we can gather the entire trace, not just the debug ones.

So hopefully the above answers the question of resource consumption by this feature.

To summarize:
- Just by logging spans (not this PR), we are consuming resources O(num_spans * num_TXes)
- This PR only consumes resources O(num_spans)

velvia:
@longbowlu @andll have addressed both of your comments.

I need to add a test and maybe refactor some of the tests.  For some reason the code coverage claims a bunch of code is not tested even though I explicitly added tests for it.

longbowlu (approved):

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -164,6 +171,11 @@ impl TelemetryConfig {
         self
     }
 
+    pub fn with_prom_registry(mut self, registry: &'reg prometheus::Registry) -> Self {
+        self.prom_registry = Some(registry);
+        self
+    }
```
huitseeker: The clone of a prometheus registry is the clone of an internal Arc. It is better to clone the registry here rather than take the lifetime generic imposed on your struct by taking a reference.
https://docs.rs/prometheus/latest/src/prometheus/registry.rs.html#220

huitseeker (commented):
I think you want to clone the registry rather than take a lifetime. Otherwise, this LGTM!
Don't worry about the code coverage, it's informative, and on your use case, a bit wonky.

bmwill (changes requested):
I'm actually not sure how i feel about this change for a few reasons:
1. This will greatly complicate the process of initializing the logging infra since the logging infra is initialized at the beginning of main while the metrics are started up sometime later inside the SuiNode constructor
2. How does this interact with testing, you can only init logging once and for testing purposes we init N nodes inside the same process. This makes it impossible to do that cleanly
3. The cardinality of that metrics may result in really poor performance on the prometheus server since there are going to be a LOT of spans in our code as well as third-party deps
4. Span timing info is already available if we have Span close events enabled. This will enable us to see how long a span was open for and could be translated into a histogram at a later point in time by looking at the logs if we'd like to.

velvia:
@bmwill 

> This will greatly complicate the process of initializing the logging infra since the logging infra is initialized at the beginning of main while the metrics are started up sometime later inside the SuiNode constructor

I've looked at that and I don't think it will be a problem to initialize the Prom registry earlier.

> How does this interact with testing, you can only init logging once and for testing purposes we init N nodes inside the same process. This makes it impossible to do that cleanly

This is optional.  We won't be starting with this option for testing, it will only be in sui-node.

> The cardinality of that metrics may result in really poor performance on the prometheus server since there are going to be a LOT of spans in our code as well as third-party deps

You have to remember that this adds a metric for each unique span in the code, at the current logging level.  At debug, we would only have like 15 spans in Sui, maybe another 15 in Narwhal.    At info, we would have none.

Prometheus can easily handle tens of thousands of histograms at the same time.   I know Prom really really well and I really do not think this will be a problem anytime soon.   

We also can filter spans as well as do sampling if we really want to, and I alluded to these elsewhere.   I don't think we will need that for a long time.

> Span timing info is already available if we have Span close events enabled. This will enable us to see how long a span was open for and could be translated into a histogram at a later point in time by looking at the logs if we'd like to.

This is true and I alluded to that.  We could take advantage of the close events timing info if we want.  

The point of this change is to enable latency insights which are rapidly actionable, and the Prometheus/Grafana ecosystem allows us to do that.   Yes, the timings are in log spans, but nobody is going to bother to go through, scrape that stuff, and built tooling to enable easy insights.   Furthermore, debug logs are very expensive and when we scale up, we will probably want to turn off all debug logging, and we won't then be able to scrape anything.

Using metrics is comparatively cheap compared to logging, and the insights are much faster - and there is an avenue to continue to measure this at scale.

bmwill (approved):

bmwill:
> I've looked at that and I don't think it will be a problem to initialize the Prom registry earlier.

I would have thought this was more difficult, glad you don't think it'll be a problem.

> This is optional. We won't be starting with this option for testing, it will only be in sui-node.

Good point, given its optional it may not be an issue.


> Prometheus can easily handle tens of thousands of histograms at the same time.

I was more worried about spans that we haven't defined ourselves, but again maybe i'm being over cautious or over estimating the number of spans.

PR #132, opened 2022-08-25
fix: reduce verbosity of read_size_from_env

huitseeker:
The message looks like a fatal error to most, whereas the condition it occurs in is perfectly innocuous. And unset variable will fallback to a default.


ebmifa (approved):

huitseeker:
Rebase error, this was actually fixed in https://github.com/MystenLabs/mysten-infra/pull/130

PR #134, opened 2022-08-30
Import pared-down version of parity-util-mem into mysten-infra

aschran:
* Removed all custom allocator features, leaving only estimate-heapsize.
* Added MallocSizeOf implementaitons for some types used in narwhal.

### DIFF
--- a/crates/parity-util-mem/src/allocators.rs
+++ b/crates/parity-util-mem/src/allocators.rs
@@ -1,0 +1,96 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright 2020 Parity Technologies
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#[cfg(feature = "std")]
+use crate::malloc_size::MallocUnconditionalSizeOf;
+use crate::malloc_size::{MallocSizeOf, MallocSizeOfOps, VoidPtrToSizeFn};
+#[cfg(not(feature = "std"))]
+use core::ffi::c_void;
+#[cfg(feature = "std")]
+use std::os::raw::c_void;
+
+mod usable_size {
+
+    use super::*;
+
+    cfg_if::cfg_if! {
+
+        if #[cfg(any(
+            target_arch = "wasm32",
+            feature = "estimate-heapsize",
+        ))] {
+
+            // do not try system allocator
+
+            /// Warning this is for compatibility only.
+            /// This function does panic: `estimate-heapsize` feature needs to be activated
+            /// to avoid this function call.
+            pub unsafe extern "C" fn malloc_usable_size(_ptr: *const c_void) -> usize {
+                unreachable!("estimate heapsize only")
+            }
+
+        } else if #[cfg(any(
+            target_os = "linux",
+            target_os = "android",
+            target_os = "freebsd",
+        ))] {
+            // Linux/BSD call system allocator (currently malloc).
+            extern "C" {
+                pub fn malloc_usable_size(ptr: *const c_void) -> usize;
+            }
+
+        } else {
+            // default allocator for non linux or windows system use estimate
+            pub unsafe extern "C" fn malloc_usable_size(_ptr: *const c_void) -> usize {
+                unreachable!("estimate heapsize or feature allocator needed")
+            }
+
+        }
+
+    }
+
+    /// No enclosing function defined.
+    #[inline]
+    pub fn new_enclosing_size_fn() -> Option<VoidPtrToSizeFn> {
+        None
+    }
+}
+
+/// Get a new instance of a MallocSizeOfOps
+pub fn new_malloc_size_ops() -> MallocSizeOfOps {
+    MallocSizeOfOps::new(
+        usable_size::malloc_usable_size,
+        usable_size::new_enclosing_size_fn(),
+        None,
+    )
+}
+
+/// Extension methods for `MallocSizeOf` trait, do not implement
+/// directly.
+/// It allows getting heapsize without exposing `MallocSizeOfOps`
+/// (a single default `MallocSizeOfOps` is used for each call).
+pub trait MallocSizeOfExt: MallocSizeOf {
+    /// Method to launch a heapsize measurement with a
+    /// fresh state.
+    fn malloc_size_of(&self) -> usize {
+        let mut ops = new_malloc_size_ops();
+        <Self as MallocSizeOf>::size_of(self, &mut ops)
+    }
+}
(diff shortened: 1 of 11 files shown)


aschran:
There are a lot of lint/clippy errors. I don't think we should actually fix those errors because it will be easier to merge upstream commits if I leave the unchanged code as-is. Do you agree, or do you think it will be too hard/annoying for us to selectively disable the CI checks on this code?

crates/parity-util-mem/derive/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
```
huitseeker: Do we need this file?

crates/parity-util-mem/derive/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! A crate for deriving the MallocSizeOf trait.
+//!
+//! This is a copy of Servo malloc_size_of_derive code, modified to work with
+//! our `parity_util_mem` library
+
```
huitseeker: Here you can add `#![allow(clippy::all)]` to silence clippy.

crates/parity-util-mem/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Crate for parity memory management related utilities.
+//! It includes global allocator choice, heap measurement and
+//! memory erasure.
+
```
huitseeker, line 15: `#![allow(clippy::all)]` would silence clippy in the whole crate

bmwill:
It looks like you're forking an existing project. Would it make more sense to have that forked into a standalone repo so that we can better sync from upstream?

aschran:
> It looks like you're forking an existing project. Would it make more sense to have that forked into a standalone repo so that we can better sync from upstream?

I don't have a strong opinion on this. François suggested just copy-pasting it in here and then deleting all of the allocator features we don't want. I'm happy to follow whatever is the normal convention for something like this? Forking into a standalone repo might be nicer in that we wouldn't have all the issues with the CI config in this repo that doesn't necessarily apply to this imported code.

@huitseeker what do you think?

huitseeker (commented):
0. The justifications to fork here rather than another repo is we want fast trait implementations for our structures that may not make sense upstream (the `IndexMap` is one posterchild), and that will be subject to the orphan rule.
1. I think the changes (1x/3 months) are not going to be that hard to integrate.
2. I think we'll need to run *some* tooling on this code. Clippy may be avoided (I've left indications on how to deactivate inline), but we'll need at least a rustfmt run.
3. We need a README at the root to explain why we need this, and how we intend to use it, and ideally a simple code example in a sibling directory of src. I know we have draft PRs using this downstream, but I'm thinking of catering to folks that don't have this context.
4. Otherwise, this LGTM!

crates/parity-util-mem/derive/CHANGELOG.md
aschran: No, I just kept it because I made zero modifications to the derive crate. But I can still scrap it if we are closing the door on trying to keep upstream merging easier (which I think we are, if we run rustfmt, etc, and your reasoning for this makes sense to me)

crates/parity-util-mem/derive/lib.rs
aschran: Thanks, done

aschran:
> 3. We need a README at the root to explain why we need this, and how we intend to use it, and ideally a simple code example in a sibling directory of src. I know we have draft PRs using this downstream, but I'm thinking of catering to folks that don't have this context.

My plan was to finish this and the PoC PR in narwhal, then come back and add a README guide at that point when all the details have been sorted out by virtue of having completed the "practice" implementation. Does that sound ok to you? I can create a first README draft immediately if you think this would be bad to initially merge without one.

aschran:
@huitseeker is there a way to skip the `crates-in-crates-directory` check or should I restructure the layout of the imported code to pass it?

bmwill:
> @huitseeker is there a way to skip the `crates-in-crates-directory` check or should I restructure the layout of the imported code to pass it?

that check doesn't have a way to disable it for a particular case (which is intentional :) )

crates/parity-util-mem/src/external_impls.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+impl<K: MallocSizeOf, V: MallocSizeOf, S> MallocSizeOf for indexmap::IndexMap<K, V, S> {
+    // This only produces a rough estimate of IndexMap size, because we cannot access private
+    // fields to measure them precisely.
+    fn size_of(&self, ops: &mut crate::MallocSizeOfOps) -> usize {
+        let mut n = self.shallow_size_of(ops);
+        if let (Some(k), Some(v)) = (K::constant_size(), V::constant_size()) {
+            n += self.len() * (k + v)
+        } else {
+            n = self
```
lanvidr:
```suggestion
            n += self
```

lanvidr (approved):
It may be more simple to keep this formatted as the rest of our code, and not worry about merging back upstream. Pending the += fix, lgtm!

crates/parity-util-mem/src/external_impls.rs
aschran: Thank you for catching this!

huitseeker:
> But I can still scrap it if we are closing the door on trying to keep upstream merging easier 

I think we are OK with scrapping it. The orphan rule will force us to put every implementation of the trait for a struct that's not accessible to us in this crate. "Easy upstream merges" is near-impossible anyway.

> is there a way to skip the crates-in-crates-directory check or should I restructure the layout of the imported code to pass it?
We should rename the directory, I think.

crates/parity-util-mem/src/allocators.rs
```diff
@@ -0,0 +1,114 @@
...
+                if HeapValidate(heap, 0, ptr as *const winapi_c_void) == 0 {
+                    ptr = *(ptr as *const *const c_void).offset(-1);
+                }
+
+                HeapSize(heap, 0, ptr as *const winapi_c_void) as usize
+            }
+
+        } else if #[cfg(any(
+            target_os = "linux",
+            target_os = "android",
```
oxade, line 42: Surely we don't need this? :)

aschran: Oh yeah, I tried to remove all that crap that I could find but thanks for pointing this out.

aschran: Done although it doesn't seem to be related to the windows rocksdb build failure.

(I made `estimate-heapsize` enabled by default so I think that should never actually be selected in any of our builds)

PR #139, opened 2022-09-02
Update to Tonic v0.8.0

macalinao:
Needed to get off of the Tonic git dependency


huitseeker (approved):
Thanks for the change! LGTM, I wouldn't mind a bump to respectively tonic 0.8.1 and tonic-health 0.7.1 while we're at it.

PR #143, opened 2022-09-06
Rename crates in a way that makes their content predictable and publishable

huitseeker:
- parity -> mysten,
- proc-macro crates -> derive


bmwill (approved):

aschran (approved):

oxade (approved):
!!!

PR #150, opened 2022-09-08
[serde] Add BcsSafeBTreeSet

lxfind:
Part of https://github.com/MystenLabs/sui/issues/5296

### DIFF
--- a/crates/serde-lib/src/bcs_safe_btreeset.rs
+++ b/crates/serde-lib/src/bcs_safe_btreeset.rs
@@ -1,0 +1,59 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+use std::collections::{BTreeMap, BTreeSet};
+
+/// Cryptographic applications using BCS cannot use BTreeSet as it does not enforce canonicity.
+/// Here we introduce a BCS-safe version of BTreeSet that's implemented using BTreeMap under
+/// the hood.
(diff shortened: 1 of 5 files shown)


bmwill (changes requested):
I don't think this is the best way to fix this flaw in bcs since it requires that we completely use a new type. The better way would probably be to introduce a serialize/deserialize function that can use used in-place of the default btreemap via "serialize-with" by annotating our uses and lets us keep using the stdlib type directly.

lxfind:
Is it possible to override the default btreemap behavior, or developer would have to explicitly annotate any use of btreemap so that they get serialized properly?

bmwill:
We'd need to annotate each instance to get the behaviour that you want. Which is no different from requiring that you use this type instead of the stdlib type. At least with the annotation route we don't have to maintain a full duplicate type that already exists in the stdlib

crates/serde-lib/src/bcs_safe_btreeset.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+use std::collections::{BTreeMap, BTreeSet};
+
+/// Cryptographic applications using BCS cannot use BTreeSet as it does not enforce canonicity.
```
kchalkias, line 7: It's not necessarily only cryptography, determinism is the feature we're looking here and it might also be useful for consistent logs and similar applications were serialization determinism is important.

PR #160, opened 2022-09-19
Add a `ClosureMetric` for Prometheus

aschran:
Implements a `ClosureMetric` for crate `prometheus` whose value is computed at the time of collection by a provided closure.


bmwill (approved):
LGTM

PR #161, opened 2022-09-22
feat(typed-store): default implementations for `Map::multi_*`

### LINKED ISSUE #40: [typed-store] create iteration based implementations for multi_ methods
Currently the `Map` trait has not default implementations for the `multi_` methods, more specifically for the:
* `multi_get`
* `multi_insert`
* `multi_remove`

it would be beneficial to offer some default implementations based on iteration for the trait. At the moment the `DBMap` which implements the `Map` trait offers its own `atomic` implementations for the above methods. Having the iteration based variants will allow other structs to basically fallback on the default (non atomic) versions.

erwanor:
Fixes #40 by providing default implementations for `Map::{multi_get, multi_insert, multi_remove}`. 

If there are plans to provide the trait with more default implementations, then it could be judicious to have a stub `SomeMap` structure implementing `Map` so that we can write tests.

### DIFF
--- a/crates/typed-store/src/traits.rs
+++ b/crates/typed-store/src/traits.rs
@@ -61,12 +61,12 @@ where
     fn values(&'a self) -> Self::Values;
 
     /// Returns a vector of values corresponding to the keys provided.
-    fn multi_get<J>(
-        &self,
-        keys: impl IntoIterator<Item = J>,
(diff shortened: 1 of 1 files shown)


huitseeker (approved):
Beautiful, thank you so much!

PR #167, opened 2022-09-29
[Telemetry-subscribers] Per-layer/separate log and span filtering; fix Tokio Console

### LINKED ISSUE #154: Re-activate Tokio Console
Actually I'm not sure this is going to work.
I can connect with Tokio-console to a Sui-node binary compiled by these changes, and with TOKIO_CONSOLE=true env var.  However, no tasks are displayed.

I followed all the recommendations in https://docs.rs/console-subscriber/0.1.7/console_subscriber/#enabling-tokio-instrumentation

You have to enable TRACE logs for Tokio and tracing, which just spews an unacceptable amount of logs.  And even after that I still am not getting any instrumentation.  Granted I had no transactions, but I should be able to see something happening just with consensus and checkpoints, etc.

_Originally posted by @velvia in https://github.com/MystenLabs/sui/issues/4438#issuecomment-1234863624_

velvia:
Originally in telemetry-subscribers we had a global filter which was based on `RUST_LOG` parsing for setting log and span output levels.   This controlled the log and span levels for all subscribers though - not just logging, but others such as automatic span latency generation and Tokio console, and apparently breaks Tokio console.  

On the advice of the Tokio team, I have switched to per-layer filtering.  This means:
- `RUST_LOG` now just controls the log/span levels for the logging output layer only
- A separate, new `span_level` controls the level of spans which are dumped out to Jaeger and used for automatic span latency measurement.  This allows us to, for example, output INFO only logs but still have very detailed TRACE level spans for measurement and detailed tracing.

This fixes #154 and Tokio Console now works for Sui.

I have also updated README to be thorough and updated the crate level docs to match the README, and renamed env vars so they are not Mysten-specific.  The above should help others who wish to use our crate.


crates/telemetry-subscribers/src/lib.rs
```diff
@@ -181,23 +280,28 @@ impl TelemetryConfig {
             self.crash_on_panic = true
         }
 
-        if env::var("MYSTEN_TRACING").is_ok() {
+        if env::var("TS_TRACING").is_ok() {
```
bmwill: Reading this I'm not quite sure what "TS" is. i think that these are probably worse names than them being prefixed with Mysten

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -272,14 +392,12 @@ impl TelemetryConfig {
...
                 .with_writer(nb_output)
+                .with_filter(log_filter)
                 .boxed();
             layers.push(fmt_layer);
         }
 
-        tracing_subscriber::registry()
-            .with(filter)
-            .with(layers)
-            .init();
```
bmwill, line 282: I actually think that the separate "span_level" knob could be completely eliminated if we restructured the way that our subscriber layers are built. Today we stick the filter first and then all of the layers in parallel. instead we could do something like:

```
tracing_subscriber::registry()
            .with(tokio_console) // if enabled
            .with(filter)
            .with(layers)
```

That way the filter is applied after tokio console.

crates/telemetry-subscribers/src/lib.rs
velvia: TS = Telemetry-subscribers (the name of this module).  I could just remove that prefix though, just like "TOKIO_CONSOLE" and "RUST_LOG"

crates/telemetry-subscribers/src/lib.rs
velvia: We could do that, but the point here is that we _want_ a separate knob.  We don't want the same knob for both logs and span-related stuff, that doesn't work at all.  For example we want automatic span latencies and tracing for spans down to trace level (if tracing is turned on), but we definitely don't want to log at TRACE level.

crates/telemetry-subscribers/src/lib.rs
velvia: Ok I renamed a bunch of the env vars to be more readable

huitseeker (commented):
This LGTM, thanks @velvia !

huitseeker (approved):

PR #168, opened 2022-09-30
fix(ci): make nightly availablein publish

huitseeker:


velvia (approved):
Hmmm... how does this interact with `rust-toolchain` file?

PR #172, opened 2022-10-06
Add prometheus metrics for tracking db operations

sadhansood:
We don't have any metrics around our db operations. This PR helps fix that to some extent.

### DIFF
--- a/crates/typed-store/src/metrics.rs
+++ b/crates/typed-store/src/metrics.rs
@@ -1,0 +1,126 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use std::sync::Arc;
+
+use once_cell::sync::OnceCell;
+use prometheus::{
+    exponential_buckets, register_histogram_vec_with_registry,
+    register_int_counter_vec_with_registry, HistogramVec, IntCounterVec, Registry,
+};
+
+#[derive(Debug)]
+pub struct DBMetrics {
+    pub rocksdb_iter_latency_seconds: HistogramVec,
+    pub rocksdb_iter_bytes: HistogramVec,
+    pub rocksdb_get_latency_seconds: HistogramVec,
+    pub rocksdb_get_bytes: HistogramVec,
+    pub rocksdb_multiget_latency_seconds: HistogramVec,
+    pub rocksdb_multiget_bytes: HistogramVec,
+    pub rocksdb_put_latency_seconds: HistogramVec,
+    pub rocksdb_put_bytes: HistogramVec,
+    pub rocksdb_delete_latency_seconds: HistogramVec,
+    pub rocksdb_deletes: IntCounterVec,
+    pub rocksdb_batch_commit_latency_seconds: HistogramVec,
+    pub rocksdb_batch_commit_bytes: HistogramVec,
+}
+
+impl DBMetrics {
+    pub fn make_db_metrics(registry: &Registry) -> &'static Arc<DBMetrics> {
+        static ONCE: OnceCell<Arc<DBMetrics>> = OnceCell::new();
+        ONCE.get_or_init(|| Arc::new(DBMetrics::new(registry)))
+    }
+    pub(crate) fn new(registry: &Registry) -> Self {
+        DBMetrics {
+            rocksdb_iter_latency_seconds: register_histogram_vec_with_registry!(
+                "rocksdb_iter_latency_seconds",
+                "Rocksdb iter latency in seconds",
+                &["cf_name"],
+                exponential_buckets(1e-6, 2.0, 24).unwrap(),
+                registry,
+            )
+            .unwrap(),
+            rocksdb_iter_bytes: register_histogram_vec_with_registry!(
+                "rocksdb_iter_bytes",
+                "Rocksdb iter size in bytes",
+                &["cf_name"],
+                registry,
+            )
+            .unwrap(),
+            rocksdb_get_latency_seconds: register_histogram_vec_with_registry!(
+                "rocksdb_get_latency_seconds",
+                "Rocksdb get latency in seconds",
+                &["cf_name"],
+                exponential_buckets(1e-6, 2.0, 24).unwrap(),
+                registry,
+            )
+            .unwrap(),
+            rocksdb_get_bytes: register_histogram_vec_with_registry!(
+                "rocksdb_get_bytes",
+                "Rocksdb get call returned data size in bytes",
+                &["cf_name"],
+                registry
+            )
+            .unwrap(),
+            rocksdb_multiget_latency_seconds: register_histogram_vec_with_registry!(
+                "rocksdb_multiget_latency_seconds",
+                "Rocksdb multiget latency in seconds",
+                &["cf_name"],
+                exponential_buckets(1e-6, 2.0, 24).unwrap(),
+                registry,
+            )
+            .unwrap(),
+            rocksdb_multiget_bytes: register_histogram_vec_with_registry!(
+                "rocksdb_multiget_bytes",
+                "Rocksdb multiget call returned data size in bytes",
+                &["cf_name"],
+                registry,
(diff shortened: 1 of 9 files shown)


crates/typed-store/src/metrics.rs
```diff
@@ -0,0 +1,126 @@
...
+        static ONCE: OnceCell<Arc<DBMetrics>> = OnceCell::new();
+        ONCE.get_or_init(|| Arc::new(DBMetrics::new(registry)))
+    }
+    pub(crate) fn new(registry: &Registry) -> Self {
+        DBMetrics {
+            rocksdb_iter_latency_seconds: register_histogram_vec_with_registry!(
+                "rocksdb_iter_latency_seconds",
+                "Rocksdb iter latency in seconds",
+                &["cf_name"],
+                exponential_buckets(1e-6, 2.0, 24).unwrap(),
```
velvia, line 38: I have some code in `span_latency_prom.rs` that calculates the factor given the top and bottom latencies, I think that's a bit clearer, feel free to steal that and refactor into a common function

crates/typed-store/src/rocks/iter.rs
```diff
@@ -32,15 +42,28 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iterator for Iter<'a, K, V> {
...
+                .db_iter
+                .key()
+                .expect("Valid iterator failed to get key");
+            let raw_value = self
                 .db_iter
                 .value()
-                .and_then(|v| bincode::deserialize(v).ok());
-
+                .expect("Valid iterator failed to get value");
+            self.db_metrics
```
velvia, line 61: I wonder how expensive it would be to monitor every write.  I think this is valuable, although write byte size can be obtained after the fact by going through a database offline.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -194,23 +214,44 @@ impl<K, V> DBMap<K, V> {
...
     }
 
     /// Consume the batch and write its operations to the database
     #[instrument(level = "trace", skip_all, err)]
     pub fn write(self) -> Result<(), TypedStoreError> {
+        let db_name = self
+            .rocksdb
+            .path()
+            .file_name()
+            .and_then(|f| f.to_str())
```
velvia, line 242: I'm worried we are allocating here for every batch  write.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -194,23 +214,44 @@ impl<K, V> DBMap<K, V> {
...
+            .unwrap_or("unknown");
+        let _timer = self
```
velvia, line 244: The timer here is unnecessary.  We already have automatic span timing measurement for any function that is instrumented with `#[intrument` ... ie each span.  It should be available today.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -298,8 +339,17 @@ where
 
     #[instrument(level = "trace", skip_all, err)]
     fn get(&self, key: &K) -> Result<Option<V>, TypedStoreError> {
+        let _timer = self
```
velvia, line 342: Again the timer here is unnecessary.  We should be able to get that today

crates/typed-store/src/rocks/mod.rs
```diff
@@ -308,8 +358,17 @@ where
 
     #[instrument(level = "trace", skip_all, err)]
     fn get_raw_bytes(&self, key: &K) -> Result<Option<Vec<u8>>, TypedStoreError> {
+        let _timer = self
```
velvia, line 361: Again the timer here is probably unnecessary.  Although, we can rename the span name to make it more unique.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -318,17 +377,33 @@ where
 
     #[instrument(level = "trace", skip_all, err)]
     fn insert(&self, key: &K, value: &V) -> Result<(), TypedStoreError> {
+        let _timer = self
```
velvia, line 380: We don't need the timer, but you can name the span better.  IE instead of adding a latency measurement, change the instrument above to say: 

    #[instrument(level = "trace", name = "rocksdb_insert", .... )]

crates/typed-store/src/rocks/mod.rs
```diff
@@ -318,17 +377,33 @@ where
...
+            .rocksdb_put_bytes
+            .with_label_values(&[&self.cf])
+            .observe((key_buf.len() + value_buf.len()) as f64);
         self.rocksdb.put_cf(&self.cf(), &key_buf, &value_buf)?;
         Ok(())
     }
 
     #[instrument(level = "trace", skip_all, err)]
     fn remove(&self, key: &K) -> Result<(), TypedStoreError> {
+        let _timer = self
```
velvia, line 397: `#[instrument(name = "rocksdb_remove", level = "trace", skip_all, err)]`

crates/typed-store/src/rocks/mod.rs
```diff
@@ -375,6 +450,11 @@ where
     where
         J: Borrow<K>,
     {
+        let _timer = self
```
velvia, line 453: Again this latency measurement is not needed because we have instrument

velvia (changes requested):
Main feedback is that we don't actually need the latency histograms because we should already have them from automatic span latency measurement layer I added.

sadhansood:
> Main feedback is that we don't actually need the latency histograms because we should already have them from automatic span latency measurement layer I added.

Just wanted to point out - we need these metrics per db and column family which seems not possible with spans, etc. One of the main goal is storage and performance tracking per cf so we can better understand our usage

crates/typed-store/src/metrics.rs
sadhansood: I'll definitely check it out

crates/typed-store/src/rocks/iter.rs
sadhansood: I think the value of this approach is the histogram of the byte sizes in real time which may be valuable in catching some tail latency kind of issues. I don't know if we should worry about perf for updating metrics in this path (there is so much else we are doing here i.e. serialization/deserialization that updating an in memory counter would dwarf in comparison in terms of cpu)

crates/typed-store/src/rocks/mod.rs
sadhansood: ok i'll move this up to the struct

crates/typed-store/src/rocks/mod.rs
sadhansood: i think i mentioned before, but this tracks metrics by db/column family which is useful and not possible with spans

crates/typed-store/src/rocks/mod.rs
sadhansood: i think i mentioned before, but this tracks metrics by db/column family which is useful and not possible with spans

crates/typed-store/src/rocks/mod.rs
sadhansood: same

crates/typed-store/src/rocks/mod.rs
sadhansood: same reasoning

crates/typed-store/src/rocks/iter.rs
velvia: That is true that serialization is going to dominate

crates/typed-store/src/rocks/mod.rs
velvia: Ok, that's fine, then go ahead.  There is actually a way to add metadata to spans such as the db_name, see 
https://docs.rs/tracing/0.1.36/tracing/attr.instrument.html
however it is not exposed yet to Prometheus as a label.

velvia (approved):
changed to approve since trace latencies don't have DB names...

PR #174, opened 2022-10-10
Blocking write

asonnino:
Add a synchronous write function to the store. 

It allows callers to know when their write operate is successful. It is used to store critical information and can be seend as the equivalent of callinf
```rust
store.write().await
store.read_notify().await
```


asonnino:
This PR fixes this issue: https://github.com/MystenLabs/sui/issues/5048

andll:
Should we instead rename `write` -> `write_async`, and then have new `blocking_write` -> `write`?

This probably will require more work on the narwhal side but going forward this will be more natural. I am pretty sure most of people who see `write` will be confused that the write is async

asonnino:
> Should we instead rename `write` -> `write_async`, and then have new `blocking_write` -> `write`?
> 
> This probably will require more work on the narwhal side but going forward this will be more natural. I am pretty sure most of people who see `write` will be confused that the write is async

I agree. I wanted to avoid introducing big changes, but if we are ready to got that way shouldn't we be extra clear an have `blocking_write` and `async_write`?

andll:
Perhaps something like `sync_write` rather than `blocking_` makes more sense than? No strong opinion though

andll (approved):
Nice! I think using sync_write is now strictly more prefer then write+notify_read, since it returns an error!

andll:
Btw, @asonnino after merging this I assume next step you are going to do is to update mysten-infra pointer in SUI and then decide in which places of nw codebase do we need sync or async write? (We should try to do this sooner rather then later since this commit basically breaks mysten-infra and sui compatibility)

asonnino:
Yes I was waiting for the new version of the repo to be published

PR #176, opened 2022-10-13
Publish new version of typed store and typed store derive

sadhansood:


velvia (commented):
Wait, can you please also upgrade `rust-toolchain` to 1.64?   
Please see https://github.com/obi1kenobi/cargo-semver-checks-action/issues/13 for more details -- it is necessary to ensure that the publish step, which includes cargo-semver-check, works correctly going forward.  I was going to do this but forgot.  Thanks!

huitseeker (commented):
Publish worked well enough in https://github.com/MystenLabs/mysten-infra/commit/2ab9a226fef02f4ecd7edb29774ad93bb6980688

Did we add any breaking change to `typed-store`?  If not, would `0.1.1` fit better?

sadhansood:
I think it is a breaking change because we now must pass in a prometheus registry during the DBMap<> initialization

velvia:
@huitseeker please see https://github.com/obi1kenobi/cargo-semver-checks-action/issues/13#issuecomment-1264208643

In short, we cannot keep relying on nightly.  The author of cargo-semver-check highly recommends we upgrade to 1.64, so we should do it.

sadhansood:
Updated the toolchain, although I am less caught up on cargo tooling it seems upgrading it should not break anything (unless someone knows otherwise)

huitseeker (approved):
Thanks @sadhansood

PR #177, opened 2022-10-14
Use same version of fastcrypto as sui

jonas-lj:
In order to align the fastcrypto version, Sui currently uses the head of this branch as the mysten-util-mem dependency. The versions here and in Sui have to be the same because MallocSizeOf is implemented here for some fastcrypto structs and used in Sui.


crates/mysten-util-mem/Cargo.toml
```diff
@@ -14,7 +14,7 @@ cfg-if = "1.0.0"
 hashbrown = { version = "0.12", optional = true }
 mysten-util-mem-derive = { path = "../mysten-util-mem-derive", version = "0.1" }
 impl-trait-for-tuples = "0.2.0"
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "4b7a06fe393003013489990dc2f4d3d2ebb32539" }
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "ab44e914084bd2fa65e7caf64ab32d61910a5d30" }
```
huitseeker, line 17: This does not match https://github.com/MystenLabs/sui/blob/127e0929b0317ee40a4f0cd95825db25616ca1d5/Cargo.toml#L96 am I missing something?

benr-ml: yes, I plan to update sui as well - https://github.com/MystenLabs/sui/pull/5533

jonas-lj: That's true. This PR will be closed, and we will merge https://github.com/MystenLabs/sui/pull/5533 instead.

PR #179, opened 2022-10-17
Add sampling for metrics reporting

sadhansood:
We should have some sampling to not degrade perf by collecting (by enabling perf context mode in rocksdb) and updating prometheus metrics for dozens of rocksdb perf context counters from every thread. 
This PR allows for configurable sampling and default is 1 sample per second per DBMap.

### DIFF
--- a/crates/typed-store-derive/Cargo.toml
+++ b/crates/typed-store-derive/Cargo.toml
@@ -15,9 +15,11 @@ proc-macro = true
 proc-macro2 = "1.0.47"
 quote = "1.0.9"
 syn = { version = "1.0.102", features = ["derive"] }
+tokio = { version = "1.21.2", features = ["full"] }
 
 [dev-dependencies]
 eyre = "0.6.8"
 rocksdb = { version = "0.19.0", features = ["snappy", "lz4", "zstd", "zlib", "multi-threaded-cf"], default-features = false }
 tempfile = "3.3.0"
 typed-store = { path = "../typed-store" }
+tokio = { version = "1.21.2", features = ["test-util"] }
--- a/crates/typed-store/src/metrics.rs
+++ b/crates/typed-store/src/metrics.rs
@@ -1,12 +1,60 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
-use std::sync::Arc;
-
 use once_cell::sync::OnceCell;
 use prometheus::{
     exponential_buckets, register_histogram_vec_with_registry,
     register_int_counter_vec_with_registry, HistogramVec, IntCounterVec, Registry,
 };
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+
+#[derive(Debug, Clone)]
+// A struct for sampling based on number of operations or duration.
+// Sampling happens if the duration expires and after number of operations
+pub struct SamplingInterval {
+    // Sample once every time duration
+    pub once_every_duration: Duration,
+    // Sample once every number of operations
+    pub after_num_ops: u64,
+    // Counter for keeping track of previous sample
+    pub counter: Arc<AtomicU64>,
+}
+
+impl Default for SamplingInterval {
+    fn default() -> Self {
+        SamplingInterval::new(Duration::from_secs(1), 0)
+    }
+}
+
+impl SamplingInterval {
+    pub fn new(once_every_duration: Duration, after_num_ops: u64) -> Self {
+        let counter = Arc::new(AtomicU64::new(1));
+        if !once_every_duration.is_zero() {
+            let counter = counter.clone();
+            tokio::task::spawn(async move {
+                loop {
+                    if counter.load(Ordering::SeqCst) > after_num_ops {
+                        counter.store(0, Ordering::SeqCst);
+                    }
+                    tokio::time::sleep(once_every_duration).await;
+                }
+            });
+        }
+        SamplingInterval {
+            once_every_duration,
+            after_num_ops,
+            counter,
+        }
+    }
+    pub fn sample(&self) -> bool {
+        if self.once_every_duration.is_zero() {
+            self.counter.fetch_add(1, Ordering::Relaxed) % (self.after_num_ops + 1) == 0
+        } else {
+            self.counter.fetch_add(1, Ordering::Relaxed) == 0
+        }
+    }
+}
 
 #[derive(Debug)]
 pub struct DBMetrics {
--- a/crates/typed-store/src/rocks/iter.rs
+++ b/crates/typed-store/src/rocks/iter.rs
@@ -5,7 +5,7 @@ use std::{marker::PhantomData, sync::Arc};
 use bincode::Options;
 use rocksdb::Direction;
 
-use crate::metrics::DBMetrics;
+use crate::metrics::{DBMetrics, SamplingInterval};
 
 use super::{be_fix_int_ser, errors::TypedStoreError};
 use serde::{de::DeserializeOwned, Serialize};
@@ -19,6 +19,7 @@ pub struct Iter<'a, K, V> {
     direction: Direction,
     cf: String,
     db_metrics: Arc<DBMetrics>,
+    read_sample_interval: SamplingInterval,
 }
 
 impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iter<'a, K, V> {
@@ -26,6 +27,7 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iter<'a, K, V> {
         db_iter: DBRawIteratorMultiThreaded<'a>,
         cf: String,
         db_metrics: &Arc<DBMetrics>,
+        read_sample_interval: &SamplingInterval,
     ) -> Self {
         Self {
             db_iter,
@@ -33,6 +35,7 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iter<'a, K, V> {
             direction: Direction::Forward,
             cf,
             db_metrics: db_metrics.clone(),
+            read_sample_interval: read_sample_interval.clone(),
         }
     }
 }
@@ -42,11 +45,16 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iterator for Iter<'a, K, V> {
 
     fn next(&mut self) -> Option<Self::Item> {
         if self.db_iter.valid() {
-            let _timer = self
-                .db_metrics
-                .rocksdb_iter_latency_seconds
-                .with_label_values(&[&self.cf])
-                .start_timer();
+            let report_metrics = if self.read_sample_interval.sample() {
+                let timer = self
+                    .db_metrics
+                    .rocksdb_iter_latency_seconds
+                    .with_label_values(&[&self.cf])
+                    .start_timer();
+                Some(timer)
+            } else {
+                None
+            };
             let config = bincode::DefaultOptions::new()
                 .with_big_endian()
                 .with_fixint_encoding();
@@ -58,12 +66,14 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iterator for Iter<'a, K, V> {
                 .db_iter
                 .value()
                 .expect("Valid iterator failed to get value");
-            self.db_metrics
-                .rocksdb_iter_bytes
-                .with_label_values(&[&self.cf])
-                .observe((raw_key.len() + raw_value.len()) as f64);
             let key = config.deserialize(raw_key).ok();
             let value = bincode::deserialize(raw_value).ok();
+            if report_metrics.is_some() {
+                self.db_metrics
+                    .rocksdb_iter_bytes
+                    .with_label_values(&[&self.cf])
+                    .observe((raw_key.len() + raw_value.len()) as f64);
+            }
             match self.direction {
                 Direction::Forward => self.db_iter.next(),
                 Direction::Reverse => self.db_iter.prev(),
--- a/crates/typed-store/tests/macro_tests.rs
+++ b/crates/typed-store/tests/macro_tests.rs
@@ -8,6 +8,8 @@ use serde::Serialize;
 use std::collections::HashSet;
 use std::fmt::Debug;
 use std::sync::Mutex;
+use std::time::Duration;
+use typed_store::metrics::SamplingInterval;
 use typed_store::rocks::list_tables;
 use typed_store::rocks::DBMap;
 use typed_store::traits::Map;
@@ -258,4 +260,36 @@ async fn store_iter_and_filter_successfully() {
         }
     }
     assert_eq!(output.len(), key_values.len());
+}
+
+#[tokio::test]
+async fn test_sampling() {
+    let sampling_interval = SamplingInterval::new(Duration::ZERO, 10);
+    for _i in 0..10 {
+        assert!(!sampling_interval.sample());
+    }
+    assert!(sampling_interval.sample());
+    for _i in 0..10 {
+        assert!(!sampling_interval.sample());
+    }
+    assert!(sampling_interval.sample());
+}
+
+#[tokio::test(flavor = "current_thread", start_paused = true)]
+async fn test_sampling_time() {
+    let sampling_interval = SamplingInterval::new(Duration::from_secs(1), 10);
+    for _i in 0..10 {
+        assert!(!sampling_interval.sample());
+    }
+    assert!(!sampling_interval.sample());
+    tokio::time::advance(Duration::from_secs(1)).await;
+    tokio::task::yield_now().await;
+    assert!(sampling_interval.sample());
+    for _i in 0..10 {
+        assert!(!sampling_interval.sample());
+    }
+    assert!(!sampling_interval.sample());
+    tokio::time::advance(Duration::from_secs(1)).await;
+    tokio::task::yield_now().await;
+    assert!(sampling_interval.sample());
 }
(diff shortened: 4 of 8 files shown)


andll:
How do we assess effectiveness of PR like this?

Incrementing prometheus counters is essentially updating atomic integer(s)
In this PR we are doing some atomic increments and then if some check passes go to incrementing prometheus counters. So we swap some atomic increments with others. It is just not immediately obvious if this is faster, and if it is, it is not clear if this is faster by any significant amount. Are there any ways we can actually measure that we are indeed improving things in a measurable way?

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,103 @@
...
+                self.perf_sampler
+                    .latest_sample_ts
+                    .store(now.as_secs(), Ordering::SeqCst);
+                return true;
+            }
+        }
+        if self.once_every_duration == Duration::MAX {
+            return false;
+        }
+        let now = SystemTime::now()
```
andll: It is (much) better to use `Instant` rather then `SystemTime` for getting local timestamps - it is both faster and more reliable

crates/typed-store/src/rocks/iter.rs
```diff
@@ -58,10 +66,12 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iterator for Iter<'a, K, V> {
                 .db_iter
                 .value()
                 .expect("Valid iterator failed to get value");
-            self.db_metrics
-                .rocksdb_iter_bytes
-                .with_label_values(&[&self.cf])
-                .observe((raw_key.len() + raw_value.len()) as f64);
+            if let Some(_timer) = report_metrics {
```
andll: Just a quick note here - this line of code did change what `_timer` of `rocksdb_iter_latency_seconds ` measures.

It previously measured time from start function till the end.

With this change it measure time from start of the function until end of this `if` block where _timer is dropped.

I am not sure how critical this is, but just pointing out that there is this subtle change.

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,103 @@
...
+        }
+        if self.after_num_ops != u64::MAX {
+            let num_ops = self.perf_sampler.num_ops.fetch_add(1, Ordering::Relaxed);
+            if num_ops >= self.after_num_ops {
+                if num_ops > self.after_num_ops {
+                    // Multiple threads got here at the same time, only one
+                    // should be able to get the sample
+                    return false;
+                }
+                let now = SystemTime::now()
```
velvia: I've been told that `Instant::now()` is faster and much preferable for measuring elapsed time compared to SystemTime.

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,103 @@
...
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("create timestamp in timing");
+        let prev_ts = self.perf_sampler.latest_sample_ts.load(Ordering::Relaxed);
+        let ts_since_last_op = now.as_secs() - prev_ts;
+        if ts_since_last_op >= self.once_every_duration.as_secs()
+            && self
+                .perf_sampler
+                .latest_sample_ts
+                .compare_exchange_weak(prev_ts, now.as_secs(), Ordering::SeqCst, Ordering::Relaxed)
```
velvia: I think it's possible to combine the two branches you have here.   I mean, the check on number of ops elapsed and time leased, you could check both, do the compare and swap, and reset the things all at once in the same branch.  You could use the compare and swap to determine who wins.

crates/typed-store/src/rocks/iter.rs
velvia: Ah, interesting.

crates/typed-store/src/rocks/iter.rs
```diff
@@ -58,10 +66,12 @@ impl<'a, K: DeserializeOwned, V: DeserializeOwned> Iterator for Iter<'a, K, V> {
...
+                self.db_metrics
+                    .rocksdb_iter_bytes
+                    .with_label_values(&[&self.cf])
```
velvia, line 74: I don't think the iter_bytes really needs to be sampled.  This is not as expensive as measuring latency.

velvia (commented):
I think my question is kinda the same.  What is the impetus for sampling now?   Is there something to show that the latency measurement is very expensive, and that the sampling is substantially less expensive?  I'm not sure, because the sampling involves timers and multiple atomic reads and writes as well.  

I'm of the mind that this sampling implementation would be almost as expensive as the latency measurement itself.  Perhaps, if you only did really simple number of operations based sampling, instead of both ops and timer based, then the cost would go down.

sadhansood:
> How do we assess effectiveness of PR like this?
> 
> Incrementing prometheus counters is essentially updating atomic integer(s) In this PR we are doing some atomic increments and then if some check passes go to incrementing prometheus counters. So we swap some atomic increments with others. It is just not immediately obvious if this is faster, and if it is, it is not clear if this is faster by any significant amount. Are there any ways we can actually measure that we are indeed improving things in a measurable way?


> I think my question is kinda the same. What is the impetus for sampling now? Is there something to show that the latency measurement is very expensive, and that the sampling is substantially less expensive? I'm not sure, because the sampling involves timers and multiple atomic reads and writes as well.

The sampling is not for guarding against one prometheus metric update. It is meant to enable/disable rocksdb perf context which measures time spent and counters across dozens of functions and locks which would slow down the system a lot if not sampled enough. The relevant PR is https://github.com/MystenLabs/mysten-infra/pull/181 (it also increases sampling duration to 10s). I realize the diffs could have been split another way so it would be much clearer and I see how the description didn't mention anything about the stacked PRs and motivation for sampling accurately

crates/typed-store/src/metrics.rs
sadhansood: The PR further down the [stack](https://github.com/MystenLabs/mysten-infra/pull/181) actually uses coarsetime::Instant which is faster and less accurate and doesn't involve a system call. I am going to bring it up here to make it easy to review

crates/typed-store/src/rocks/iter.rs
sadhansood: ah good catch, i'll move it down after deserialize

andll: You can also just use `report_metrics.is_some()` here

crates/typed-store/src/metrics.rs
sadhansood: The PR further down this stack https://github.com/MystenLabs/mysten-infra/pull/181 actually uses coarsetime::Instant which is faster and less accurate and doesn't involve a system call. I am going to bring it up here to make it easy to review

andll:
So is the goal is to simply have measurement of some internal Rocks counters every 10 second?

If this is the case, what do you think about this alternative to using coarse time and more complex logic with counters:

* Have a single field - `counter: Arc<AtomicU64>` in the sampler
* Have a thread/tokio task that has a loop that calls `counter.set(0)` every interval. (and it can check strong count on the counter to exit when all the references to sampler has died out)
* When we need to make a decision whether to do the sampling, call `if counter.inc(SeqCst) == 0 {do measurement}`

I think this is probably going to be much easier to reason about, and it is also faster

sadhansood:
> So is the goal is to simply have measurement of some internal Rocks counters every 10 second?
> 
> If this is the case, what do you think about this alternative to using coarse time and more complex logic with counters:
> 
> * Have a single field - `counter: Arc<AtomicU64>` in the sampler
> * Have a thread/tokio task that has a loop that calls `counter.set(0)` every interval. (and it can check strong count on the counter to exit when all the references to sampler has died out)
> * When we need to make a decision whether to do the sampling, call `if counter.inc(SeqCst) == 0 {do measurement}`
> 
> I think this is probably going to be much easier to reason about, and it is also faster

The goal is to have a general purpose sampler which can sample based on counts and time duration. While writing this I realized the implementation actually is different from the intended goal (and fixing it as well). The default usage of the sampler is indeed just once per second right now but switching b/w either modes (count and/or duration based) during debugging and in future is not out of question. Also a user will be able to configure the sampling for individual DBMaps and choose the right configuration themselves in the future

crates/typed-store/src/metrics.rs
sadhansood: The reason for doing one at a time is to optimize away the checking of the other one

crates/typed-store/src/rocks/iter.rs
sadhansood: This one could actually be expensive as it is happening in the next() call which might just be iterating over data in memory.

andll:
So I would say if the current usage is to just measure something every n seconds, then I would suggest to take the simplest approach(something along the lines of what I have outlined with resetting atomic integer every n second by the timer). It is always possible to come back later and make things more complicated when the need arise.

sadhansood:
I disagree mainly because we do want to be able to do things like log 1 in every 100 get calls which failed (but not more than once per second), etc (not in this stacked list of PRs which is already big enough). If you have suggestions on how to make things simple *with the current requirements* please feel free to suggest in your review.

andll:
> log 1 in every 100 get calls which failed (but not more than once per second)
> If you have suggestions on how to make things simple with the current requirements please feel free to suggest in your review

Yes I do :)

So If the goal to log every 100 calls, but no more then once per second:

We can have a single `counter: AtomicU64`

When sampling:
```
if counter.inc() == 100 {do sampling}
```

Additional timer:
```
loop {
if counter.load() >= 100 {counter.set(0)}
sleep(1s);
}
```

I personally think that this is simpler **and** likely will be faster. But I guess if you feel strongly that current implementation makes more sense I am also ok with that

sadhansood:
> > log 1 in every 100 get calls which failed (but not more than once per second)
> > If you have suggestions on how to make things simple with the current requirements please feel free to suggest in your review
> 
> Yes I do :)
> 
> So If the goal to log every 100 calls, but no more then once per second:
> 
> We can have a single `counter: AtomicU64`
> 
> When sampling:
> 
> ```
> if counter.inc() == 100 {do sampling}
> ```
> 
> Additional timer:
> 
> ```
> loop {
> if counter.load() >= 100 {counter.set(0)}
> sleep(1s);
> }
> ```
> 
> I personally think that this is simpler **and** likely will be faster. But I guess if you feel strongly that current implementation makes more sense I am also ok with that

Cool! Let's do it this way then - see if you find anything missing now. Thanks

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,72 @@
...
+
+impl SamplingInterval {
+    pub fn new(once_every_duration: Duration, after_num_ops: u64) -> Self {
+        let counter = if !once_every_duration.is_zero() {
+            // start with counter set to 1 instead of 0 so first check of
+            // sample() does not return true
+            let counter = Arc::new(AtomicU64::new(1));
+            let counter_cloned = counter.clone();
+            tokio::task::spawn(async move {
+                let counter = counter_cloned.clone();
```
andll: Why do you clone here?

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,72 @@
...
+                loop {
+                    let counter = counter.clone();
```
andll: ...and here?

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,72 @@
...
+
+impl Default for SamplingInterval {
+    fn default() -> Self {
+        SamplingInterval::new(Duration::from_secs(1), 0)
+    }
+}
+
+impl SamplingInterval {
+    pub fn new(once_every_duration: Duration, after_num_ops: u64) -> Self {
+        let counter = if !once_every_duration.is_zero() {
```
andll: Would it be cleaner to just write this code as
```
let counter = Arc::new(AtomicU64::new(0));
if !once_every_duration.is_zero() {
/// work with this counter
}
```

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,72 @@
...
+    pub fn sample(&self) -> bool {
+        if self.once_every_duration.is_zero() {
+            if self.counter.fetch_add(1, Ordering::Relaxed) == self.after_num_ops {
+                self.counter.store(0, Ordering::SeqCst);
+                true
+            } else {
+                false
+            }
+        } else {
+            self.counter.fetch_add(0, Ordering::Relaxed) == 0
```
andll: So if `once_every_duration` is not zero you always return true in sample? (counter is initialized with 0 and never changes value?)

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,72 @@
...
+    pub counter: Arc<AtomicU64>,
+}
+
+impl Default for SamplingInterval {
+    fn default() -> Self {
+        SamplingInterval::new(Duration::from_secs(1), 0)
+    }
+}
+
+impl SamplingInterval {
```
andll, line 30: Perhaps we should also add unit test for `once_every_duration != Duration::ZERO`

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,72 @@
...
+        };
+        SamplingInterval {
+            once_every_duration,
+            after_num_ops,
+            counter,
+        }
+    }
+    pub fn sample(&self) -> bool {
+        if self.once_every_duration.is_zero() {
+            if self.counter.fetch_add(1, Ordering::Relaxed) == self.after_num_ops {
```
andll: You can also do
```
self.counter.fetch_add(1, Ordering::Relaxed)  % self.after_num_ops == 0
```
(and then don't do `self.counter.store(0)`)
This would be more precise and avoid race conditions on the counter

crates/typed-store/tests/macro_tests.rs
```diff
@@ -259,3 +261,16 @@ async fn store_iter_and_filter_successfully() {
     }
     assert_eq!(output.len(), key_values.len());
 }
+
+#[tokio::test]
+async fn test_sampling() {
```
andll, line 266: Should we also have unit test for non-zero sampling interval?
You can use tokio test with start_paused=true and manual advanced time(sui codebase have examples)

crates/typed-store/src/metrics.rs
sadhansood: counter is initialized with 1, it changes value because you do fetch_add?

crates/typed-store/src/metrics.rs
sadhansood: notice the counter to be initialized with 1 and 0 in if/else block?

crates/typed-store/tests/macro_tests.rs
sadhansood: oh yeah that's a good point, we should have that test. I;ll add it

crates/typed-store/src/metrics.rs
```diff
@@ -1,12 +1,72 @@
...
+#[derive(Debug, Clone)]
+// A struct for sampling based on number of operations or duration.
+// Sampling happens if the duration expires and after number of operations
+pub struct SamplingInterval {
+    // Sample once every time duration
+    pub once_every_duration: Duration,
+    // Sample once every number of operations
+    pub after_num_ops: u64,
+    // Counter for keeping track of previous sample
+    pub counter: Arc<AtomicU64>,
```
velvia, line 21: Here @sadhansood , you use an Arc of Atomic here for counter.  However if SamplingInterval is purely used within `ThreadLocal` then Arc of Atomic is not necessary.

I think you used to have an atomic for the time interval too but you must have cleaned that up.

crates/typed-store/src/metrics.rs
sadhansood: hmm, that could technically overflow the counter if we never reset it

crates/typed-store/src/metrics.rs
sadhansood: @velvia - SamplingInterval is not used within any threadlocal. The perf context is thread local, however we need to sample enabling perf context across multiple threads (and hence thread safe SamplingInterval is needed)

crates/typed-store/src/metrics.rs
sadhansood: Because we return the counter from the if block, we need to clone to pass it along to the task running in parallel? The compiler thinks so

crates/typed-store/src/metrics.rs
andll: But you are adding `0` with fetch_add(0), that does not change counter value right?

crates/typed-store/src/metrics.rs
andll: Yeah, this is on line 36, but why do you clone it inside the task?

crates/typed-store/src/metrics.rs
andll: I think it is not practically likely that we overflow u64 counter.

crates/typed-store/src/metrics.rs
sadhansood: oh you are so right!

andll (approved):

bmwill:
You may also want to take a look at a similar tool that I wrote for diem:
https://github.com/diem/diem/blob/main/crates/diem-logger/src/sample.rs

sadhansood:
> You may also want to take a look at a similar tool that I wrote for diem: https://github.com/diem/diem/blob/main/crates/diem-logger/src/sample.rs

It is pretty similar but it does have one difference that it does not provide sampling with both count and duration (it does work for either one of those) taken into consideration

crates/typed-store-derive/Cargo.toml
```diff
@@ -15,6 +15,7 @@ proc-macro = true
 proc-macro2 = "1.0.47"
 quote = "1.0.9"
 syn = { version = "1.0.102", features = ["derive"] }
+tokio = { version = "1.21.2", features = ["full", "test-util"] }
```
bmwill: please do not enable test-utils in normal builds.

bmwill (dismissed):

crates/typed-store-derive/Cargo.toml
bmwill: you can have another `tokio` line under the dev-deps section to enable this for testing purposes.

PR #180, opened 2022-10-17
Sadhan/add_cf_metrics

sadhansood:
This PR allows us to capture rocksdb internal metrics per column family. These metrics are updated in the background

### DIFF
--- a/crates/typed-store/src/metrics.rs
+++ b/crates/typed-store/src/metrics.rs
@@ -342,6 +342,13 @@ impl DBMetrics {
         }
     }
     pub fn make_db_metrics(registry: &Registry) -> &'static Arc<DBMetrics> {
+        // TODO: Remove static because this basically means we can
+        // only ever initialize db metrics once with a registry whereas
+        // in the code we might be trying to initialize it with different
+        // registries. The problem is underlying metrics cannot be re-initialized
+        // or prometheus complains. We essentially need metrics per column family
+        // but that might cause an explosion of metrics and hence a better way
+        // to do this is desired
         static ONCE: OnceCell<Arc<DBMetrics>> = OnceCell::new();
         ONCE.get_or_init(|| Arc::new(DBMetrics::new(registry)))
     }
--- a/crates/typed-store/src/metrics.rs
+++ b/crates/typed-store/src/metrics.rs
@@ -3,7 +3,8 @@
 use once_cell::sync::OnceCell;
 use prometheus::{
     exponential_buckets, register_histogram_vec_with_registry,
-    register_int_counter_vec_with_registry, HistogramVec, IntCounterVec, Registry,
+    register_int_counter_vec_with_registry, register_int_gauge_vec_with_registry, HistogramVec,
+    IntCounterVec, IntGaugeVec, Registry,
 };
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


crates/typed-store/src/rocks/mod.rs
```diff
@@ -172,6 +205,160 @@ impl<K, V> DBMap<K, V> {
...
+            .set(
+                Self::get_int_property(rocksdb, &cf, properties::BACKGROUND_ERRORS)
+                    .unwrap_or(METRICS_ERROR),
+            );
+    }
+}
+
+impl<K, V> Drop for DBMap<K, V> {
+    fn drop(&mut self) {
+        self.exit_signal.store(true, Ordering::SeqCst);
```
andll: I think this is not quite right when considering that DBMap has derived Clone implementation.

When we call `DBMap::new` we start the thread. After that DBMap can be cloned.

Then it seems that if any of cloned DBMap is dropped we stop the reporting thread. I think more reasonable expectation is that if **all** cloned DBMap dropped then we stop reporting.

In practice it is fairly easy to implement if exit_signal will be just `Arc<()>` inside the DBMap, and then reporting thread will verify `strong_count` on that Arc

crates/typed-store/src/rocks/mod.rs
```diff
@@ -94,11 +112,40 @@ pub struct DBMap<K, V> {
...
+        db: Arc<rocksdb::DBWithThreadMode<MultiThreaded>>,
+        opt_cf: &str,
+        db_metrics: Arc<DBMetrics>,
+    ) -> Self {
+        let exit = Arc::new(AtomicBool::new(false));
+        let exit_cloned = exit.clone();
+        let db_cloned = db.clone();
+        let db_metrics_cloned = db_metrics.clone();
+        let cf = opt_cf.to_string();
+        let _cf_metric_reporter = Arc::new(std::thread::spawn(move || {
```
andll: It seems `_cf_metric_reporter` is never used, so why have it?

crates/typed-store/src/rocks/mod.rs
```diff
@@ -94,11 +112,40 @@ pub struct DBMap<K, V> {
...
+            while !exit_cloned.load(Ordering::Relaxed) {
+                Self::report_metrics(&db, &cf, &db_metrics);
+                sleep(Duration::from_millis(CF_METRICS_REPORT_PERIOD_MILLIS));
```
andll: Assuming that `Self::report_metrics` is not too expensive in terms of CPU (it seems it just reads some variables from rocksdb), it might be better to have this reporting thread to be a tokio task on the main scheduler.

So instead of using `thread::spawn` it seem to be more reasonable to use `tokio::spawn` and then use `tokio::sleep` here. This way we will avoid creating new thread for each DBMap

crates/typed-store/src/rocks/mod.rs
sadhansood: The original intention was to invoke join on the thread handle, but looks like that never happened. I'll add the join which is what we should be doing anyways

crates/typed-store/src/rocks/mod.rs
sadhansood: This I am not 100% certain of. It is likely better to use spawn_blocking to be safe.

andll: I see. if you do want a blocking thread then `thread::spawn` probably makes more sense (spawn_blocking might have some additional overhead on thread housekeeping that we don't care about)

crates/typed-store/src/rocks/mod.rs
```diff
@@ -94,11 +113,59 @@ pub struct DBMap<K, V> {
...
+            cf: opt_cf.to_string(),
+            db_metrics: db_metrics_cloned,
+            read_sample_interval: SamplingInterval::default(),
+            write_sample_interval: SamplingInterval::default(),
+            metric_reporter: Arc::new(Mutex::new(Some(metric_reporter))),
+            exit_signal: exit_cloned,
+        }
+    }
+
+    pub async fn stop(&self) -> Result<(), TypedStoreError> {
```
andll: This is not called anywhere?

Tbh, I think this went off the initially correct corse of stopping the reporting task when the last cloned instance of DBMap is dropped, and now we ended up having this `async stop` method for which it is not clear who is going to call it.

I think it is also ok not to join on the reporting thread to finish, as I suspect that making this join work correctly can be a bit tricky

Anyway, what about following proposal:

(1) Make `exit_signal: Arc<()>` 
(2) Keep element of this `Arc<()>` in DBMap
(3) In the reporting thread have `while Arc::strong_count(exit_signal) > 1`

This way reporting thread stops when DBMap is dropped, no need to dedicated `stop` method

sadhansood: The idea is to have stop() be called in the exit routine of wherever DBMap is being used (sui, etc). The motivation is to cleanup any spawned threads before exiting the program. The problem I see with Arc::strong_count is it does things the other way and does not guarantee a proper waiting/shutdown.

crates/typed-store/Cargo.toml
```diff
@@ -20,7 +20,7 @@ prometheus = "0.13.2"
 rocksdb = { version = "0.19.0", features = ["snappy", "lz4", "zstd", "zlib", "multi-threaded-cf"], default-features = false }
 serde = { version = "1.0.140", features = ["derive"] }
 thiserror = "1.0.37"
-tokio = { version = "1.21.2", features = ["sync", "macros", "rt"] }
+tokio = { version = "1.21.2", features = ["full", "test-util"] }
```
bmwill: I would recommend against enabling test-utils in normal builds.

crates/typed-store/src/metrics.rs
```diff
@@ -172,3 +319,22 @@ impl DBMetrics {
...
+    pub(crate) fn new(registry: &Registry) -> Self {
+        DBMetrics {
+            op_metrics: OperationMetrics::new(registry),
+            cf_metrics: ColumnFamilyMetrics::new(registry),
+        }
+    }
+    pub fn make_db_metrics(registry: &Registry) -> &'static Arc<DBMetrics> {
+        static ONCE: OnceCell<Arc<DBMetrics>> = OnceCell::new();
+        ONCE.get_or_init(|| Arc::new(DBMetrics::new(registry)))
+    }
```
bmwill: This is going to present a problem since what happens if i call `make_db_metrics` twice with two different registries? (We will likely do this in a number of places). Please re-think this, or remove it.

crates/typed-store/src/rocks/mod.rs
sadhansood: The problem with this is there is no guarantee of it ever cleaning up the reporting thread. If the sleep within the loop was any longer, the main thread would drop the Arc<> but the reporting thread may never get to checking it and the program terminates with lingering threads. Having a stop() method is a guarantee that the user of this library gets that by invoking they will get the metric reported one last time before the thread terminates.

crates/typed-store/src/rocks/mod.rs
bmwill: I think it would probably be better to use a tokio task for this, if we're assuming that a DB is already always going to be constructed in a a tokio runtime. The ergonomics and capabilities are going to dramatically improve how this metrics reporter task operate.

crates/typed-store/src/rocks/mod.rs
bmwill: eg, we can select between a cancle handle and an interval and we can close the task immediately and not have to wait for a whole other round of sleeps

bmwill: The actual logic of collecting the metrics could be done via spawn_blocking if we really are worried (from inside a tokio task). I would recommend not using std::thread here.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -172,6 +205,160 @@ impl<K, V> DBMap<K, V> {
...
+            Ok(Some(value)) => Ok(value.try_into().unwrap()),
+            Ok(None) => Ok(0),
+            Err(e) => Err(TypedStoreError::RocksDBError(e.into_string())),
+        }
+    }
+
+    fn report_metrics(
+        rocksdb: &Arc<rocksdb::DBWithThreadMode<MultiThreaded>>,
+        cf_name: &str,
+        db_metrics: &Arc<DBMetrics>,
```
bmwill, line 238: change `&Arc` to just `&` to avoid needing to do too many derefs on Arcs.

bmwill (changes requested):

crates/typed-store/src/metrics.rs
sadhansood: It is a todo to find a better solution for this (unrelated to this PR btw). If we did not create a static instance, we will end up reinitializing these prometheus metrics every time we create a new DBMap which prometheus does not like and complains. The right way to fix this is to create an instance of Arc<DBMetrics> in sui initialization code and pass it down everywhere DBMap is being created through some kind of macro based options. We do only use one registry to report metrics in a sui instance though, no?

crates/typed-store/src/metrics.rs
```diff
@@ -172,3 +327,22 @@ impl DBMetrics {
...
```
bmwill: Then could you put a TODO comment here explaining the issue?

bmwill: I suppose this could be fixed by creating separate metrics for CF's instead of using CF as a dimension to the metric. But placing a comment to address should be fine for now.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -99,11 +109,52 @@ pub struct DBMap<K, V> {
     db_metrics: Arc<DBMetrics>,
     read_sample_interval: SamplingInterval,
     write_sample_interval: SamplingInterval,
+    _sender: Arc<oneshot::Sender<()>>,
```
bmwill: Can we name this something a bit more descriptive of what its purpose is? Something like `_metrics_task_cancel_handle` or even a descriptive comment of its use would also suffice.

crates/typed-store/src/metrics.rs
velvia: Yes we only use one registry.  At least for Sui.   Narwhal code used its own registry, but I'm not sure about now now that it's integrated (Narwhal has its own metrics port too, again if it was one registry it would be integrated - @bmwill do you know?)

bmwill (approved):

PR #181, opened 2022-10-18
Sadhan/perf_ctx_metrics

sadhansood:
Add rocksdb read and write perf context metrics in db library. These metrics are quite useful in debugging performance issues in real time.
Additionally, using coarsetime::Instant instead of SystemTIme in the hot path which updates timestamp in a background job

### DIFF
--- a/crates/typed-store/src/metrics.rs
+++ b/crates/typed-store/src/metrics.rs
@@ -6,9 +6,16 @@ use prometheus::{
     register_int_counter_vec_with_registry, register_int_gauge_vec_with_registry, HistogramVec,
     IntCounterVec, IntGaugeVec, Registry,
 };
+use rocksdb::perf::set_perf_stats;
+use rocksdb::{PerfContext, PerfMetric, PerfStatsLevel};
+use std::cell::RefCell;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
+
+thread_local! {
+    static PER_THREAD_ROCKS_PERF_CONTEXT: std::cell::RefCell<rocksdb::PerfContext>  = RefCell::new(PerfContext::default());
+}
 
 #[derive(Debug, Clone)]
 // A struct for sampling based on number of operations or duration.
@@ -24,7 +31,7 @@ pub struct SamplingInterval {
 
 impl Default for SamplingInterval {
     fn default() -> Self {
-        SamplingInterval::new(Duration::from_secs(1), 0)
+        SamplingInterval::new(Duration::from_secs(60), 0)
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/typed-store/src/metrics.rs
```diff
@@ -1,14 +1,20 @@
...
+use std::cell::RefCell;
 use std::sync::atomic::{AtomicU64, Ordering};
-use std::time::{Duration, UNIX_EPOCH};
-use std::{sync::Arc, time::SystemTime};
+use std::sync::Arc;
+use std::time::Duration;
+
+use rocksdb::{perf::set_perf_stats, PerfContext, PerfMetric, PerfStatsLevel};
+
+thread_local! {static PER_THREAD_ROCKS_PERF_CONTEXT: std::cell::RefCell<PerfContext>  = RefCell::new(PerfContext::default());}
```
bmwill: Would you be able to format this so that it is a bit more readable?
```suggestion
thread_local! {
    static PER_THREAD_ROCKS_PERF_CONTEXT: std::cell::RefCell<PerfContext>  = RefCell::new(PerfContext::default());
}
```

velvia (commented):
My recommendation after looking at both this PR and the sampling PR is the following:

1. Instead of using sampling on RocksDB methods which is kind of expensive, for gathering these RocksDB perf metrics, it would be better to just use Tokio scheduler to schedule a task to regularly get and update these metrics.  That way, there does not need to be a per-RocksDB call cost to what is essentially a timed gathering of perf stats.
2. Just remove the sampling from the RocksDB methods.  The bytes Len and latency measurement are not really that expensive.

sadhansood:
> 1. Instead of using sampling on RocksDB methods which is kind of expensive, for gathering these RocksDB perf metrics, it would be better to just use Tokio scheduler to schedule a task to regularly get and update these metrics

You cannot collect this in the background. The perf context is thread local which basically means it has to be enabled in the same thread which is calling the rocksdb get, put, etc

sadhansood:
> 2\. Just remove the sampling from the RocksDB methods.  The bytes Len and latency measurement are not really that expensive.

This is not very unreasonable for everything but iterator. I'll remove the sampling for op metrics

velvia:
> You cannot collect this in the background. The perf context is thread local which basically means it has to be enabled in the same thread which is calling the rocksdb get, put, etc

Ok, got it.  
Then we just have to make the thread local stuff cheaper.  I remember a whole bunch of atomic reads/writes - why are those necessary if you are using thread locals?   We just have to make each measurement cheaper then.
I'm not sure which is the most expensive part - the atomic stuff, obtaining a timer (which should be really cheap CPU register).....

sadhansood:
> > You cannot collect this in the background. The perf context is thread local which basically means it has to be enabled in the same thread which is calling the rocksdb get, put, etc
> 
> Ok, got it. Then we just have to make the thread local stuff cheaper. I remember a whole bunch of atomic reads/writes - why are those necessary if you are using thread locals? We just have to make each measurement cheaper then. I'm not sure which is the most expensive part - the atomic stuff, obtaining a timer (which should be really cheap CPU register).....

I do not think I added a whole bunch of atomic read/writes. Could you may be point me the PR? These are the other rwo PRs for quick reference https://github.com/MystenLabs/mysten-infra/pull/179 and https://github.com/MystenLabs/mysten-infra/pull/180. Also just leaving here more info on what perf context is https://github.com/aayushKumarJarvis/rocks-wiki/blob/master/Perf-Context-and-IO-Stats-Context.md and how RocksDB recommends to use it (using sampling, etc)

crates/typed-store/src/metrics.rs
```diff
@@ -1,15 +1,261 @@
...
+                    .latest_sample_ts
+                    .store(now.as_secs(), Ordering::SeqCst);
+                return true;
+            }
+        }
+        if self.once_every_duration == Duration::MAX {
+            return false;
+        }
+        let now = Clock::recent_since_epoch();
+        let prev_ts = self.perf_sampler.latest_sample_ts.load(Ordering::Relaxed);
```
velvia: Hmm, so this PerfSampler seems to have been brought over from the other PR, but it seems lots of optimizations went into this and this seems to be an earlier rev of the design?   Cuz this still has an Atomic for the prev_ts and didn't seem to have a bunch of other changes?

velvia (commented):
My only major comment is - do we need all these diff metrics or could we start with less metrics?

Also, while you're doing this could you add a gauge for the amount of memory used for each DB?   I meant to do this but you're already adding metrics.  There is already a method, `get_memory_usage()` in the typed-store-derive crate.  Thanks

crates/typed-store/src/metrics.rs
sadhansood: Just rebased, and it should look better now

sadhansood:
I think we can find after sometime which ones of these are useful and remove the ones which are not adding much value. I am happy to do metric cleanup once we have seen their value. Do you mind if I add the prometheus metric for `get_memory_usage` in a follow up diff?

velvia (approved):
I think this mostly looks fine now.
What is the expected initial sampling rate?  We shouldn't really need to sample these more than, say, once a minute right?

sadhansood:
Yes definitely, it is actually once per 10s but I think 60s makes more sense. I'll make it once a minute

PR #189, opened 2022-10-25
Fix how we init db metrics

sadhansood:
Right now the way we initialize db metrics is a little weird in the sense every db map that is created needs to pass in the prometheus registry and we return the static instance in make_db_metrics.
This PR fixes the behavior by giving user clear expectations i.e. initialize DBMetrics just once and do not think about passing in the registry anymore:
1. In the main() routine after we create the prometheus registry - we should call DBMetrics::init()
2. No need to pass in registry anywhere after that as DBMap will just use the globally initialized db metrics with DBMetrics::get()

### DIFF
--- a/crates/typed-store-derive/src/lib.rs
+++ b/crates/typed-store-derive/src/lib.rs
@@ -407,7 +407,7 @@ pub fn derive_dbmap_utils_general(input: TokenStream) -> TokenStream {
                 tables_db_options_override: Option<typed_store::rocks::DBMapTableConfigMap>
             ) -> Self {
                 let path = &path;
-                let (db, db_metrics) = {
+                let db = {
                     let opt_cfs = match tables_db_options_override {
                         None => [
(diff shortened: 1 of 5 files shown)


bmwill (changes requested):
This is not a solution to the problem. Why don't we creat a metricc type and then require that it is passed in at db initialization time?

bmwill (approved):
I'm fine with this for now, but i do think we'll need to thread through a DBMetrics type so that we can have dbs that export to different registries.

sadhansood:
> I'm fine with this for now, but i do think we'll need to thread through a DBMetrics type so that we can have dbs that export to different registries.

Even though it is a little challenge to do so (because DBMaps are scattered over different files), I agree with above - the right fix is to be able to pass in the DBMetricsType with whatever registry user wants to use. Thanks @bmwill for putting it out.

PR #192, opened 2022-10-25
Make sync use full feature

sadhansood:
cargo fails to publish this crate today because sync only uses "derive" feature, instead we should use "full". The failure is here:
https://github.com/MystenLabs/mysten-infra/actions/runs/3291600851/jobs/5494042110#step:8:79


velvia (approved):
Awesome thanks.  Note: please change title of PR if possible to `syn` instead of `sync`

PR #195, opened 2022-10-29
codec: use bincode::deserialize instead of deserialize_from

bmwill:


oxade (approved):

oxade:
I'm not 100% sure why this works better than what we have, but stamping to move things along.

